PR #2, opened 2022-04-14
01b Tokens

ngjunsiang:
# Categorising tokens

We have a list of tokens, each represented as a string. This is a start, but it won't be very much help to us in the next phase of massaging meaning out of this string of tokens.

For a start, what is a `DECLARE`? What is `i`? What does the `\n` do? It's hard to know when we treat all tokens as being the same.

Let's categorise our tokens. Instead of a simple string, let's add some categorical information about each token. For a start, let's:

- identify whether it is a keyword (including symbols), integer, string, or name
- if it's an integer or string, convert it to an appropriate Python value for easier use later.

### DIFF
--- a/scanner.py
+++ b/scanner.py
@@ -1,3 +1,9 @@
+# Errors
+
+class ParseError(Exception): pass
+
+
+
 # Token types
 
 KEYWORDS = [
@@ -100,7 +106,7 @@ def scan(src):
             text = symbol(code)
             token = makeToken('symbol', text, None)
         else:
-            raise ValueError(f"Unrecognised character {repr(char)}.")
+            raise ParseError(f"Unrecognised character {repr(char)}.")
         tokens += [token]
         print('Scanned token:', token, ', characters left:', len(code['src']))
     return tokens
--- a/scanner.py
+++ b/scanner.py
@@ -1,3 +1,22 @@
+# Token types
+
+KEYWORDS = [
+    'DECLARE', 'OUTPUT',
+    'WHILE', 'DO', 'ENDWHILE',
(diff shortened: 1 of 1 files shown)


ngjunsiang:
# Token types [aa5af6309778059bd7350fccf88a04cbf93214b0]

Make a list of valid tokens to check against.

ngjunsiang:
# Representing tokens

We'll store each token as a Python dict, to hold this information. Each token will have the following keys:

- `'type'`: represents the token's type (keyword, integer, string, name, eol)
- `'word'`: represents how the token is represented in the source code. (This is what we scanned earlier.)
- `'value'`: An appropriate Python representation of the token, if it is of integer or string type.

This looks potentially repetitive, so let's make a helper function: [701e5a6a8138a1a694c2de89abeccd0c6c9c7e47]

ngjunsiang:
# Using `makeToken()`

Update `scan()` to use the helper function and represent tokens as dicts: [7b7928da5e674eff904b347722d6663be0bf63e1]

ngjunsiang:
# Custom errors: `ParseError`

It is good practice when writing your own programs to include your own custom errors. The usefulness of this will become clearer once we "graduate" from scanning and parsing, to writing the entire program.

We make our own `ParseError`:

https://github.com/nyjc-computing/pseudo/blob/11e11dc1211c33a49ec8898ea4061d8b86ffa40a/scanner.py#L3

And then `raise` it when we find an unrecognised character:

https://github.com/nyjc-computing/pseudo/blob/11e11dc1211c33a49ec8898ea4061d8b86ffa40a/scanner.py#L108-L109

PR #3, opened 2022-04-14
Parser

ngjunsiang:
# Expressions

Before we start to get deep into the weeds, let's make a calculator first. A calculator is not an interpreter yet. It can handle the following:

`1 + 1`

A working calculator will display the result `2`. How does it get to that result? It has to have some internal representation of the operation.

**Expressions** are what we call these representations. With our scanner, we now have a sequence of tokens instead of plain characters

Test code:

```
import scanner

scanner.scan('1 + 1')
```

Result:

```
[{'type': 'integer', 'word': '1', 'value': 1},
 {'type': 'symbol', 'word': '+', 'value': None},
 {'type': 'integer', 'word': '1', 'value': 1},
 {'type': 'keyword', 'word': '\n', 'value': None}]
```

More complex expressions, such as `2 * 3 - 1 = 1 / 5 - 4`, have a different order of interpretation. To interpret the operations in the correct order, we cannot interpret this sequence of tokens linearly; we build them up into a structured expression tree. This process is called **parsing**.

### DIFF
--- a/builtin.py
+++ b/builtin.py
@@ -20,3 +20,54 @@ OPERATORS = [
 ]
 
 SYMBOLS = [':']
+
+
+
+# Operators
+
+def add(x, y):
+    return x + y
+
+def sub(x, y):
+    return x - y
+
+def mul(x, y):
+    return x * y
+
+def div(x, y):
+    return x / y
+
+def lt(x, y):
+    return x < y
+
+def lte(x, y):
+    return x <= y
+
+def gt(x, y):
+    return x > y
+
+def gte(x, y):
+    return x >= y
+
+def ne(x, y):
+    return x != y
+
+def eq(x, y):
+    return x == y
+
+
+
+# Classifier
+
+operators = {
+    '+': add,
+    '-': sub,
+    '*': mul,
+    '/': div,
+    '<': lt,
+    '<=': lte,
+    '>': gt,
+    '>=': gte,
+    '<>': ne,
+    '=': eq,
+}
--- a/builtin.py
+++ b/builtin.py
@@ -1,0 +1,3 @@
+# Errors
+
+class ParseError(Exception): pass
--- a/builtin.py
+++ b/builtin.py
@@ -1,3 +1,22 @@
 # Errors
 
 class ParseError(Exception): pass
+
+
+
+# Token types
+
+KEYWORDS = [
+    'DECLARE', 'OUTPUT',
+    'WHILE', 'DO', 'ENDWHILE',
+    'CASE', 'OF', 'OTHERWISE', 'ENDCASE',
+]
+
+TYPES = ['INTEGER', 'STRING']
+
+OPERATORS = [
+    '+', '-', '/', '*', '=',
+    '<', '<-', '<=', '>', '>=', '<>',
+]
+
+SYMBOLS = [':']
--- a/parser.py
+++ b/parser.py
@@ -35,7 +35,14 @@ def eq(x, y):
 # Classifier
 
 operators = {
-    '+': add, '-': sub, '*': mul, '/': div,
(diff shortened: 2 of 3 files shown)


ngjunsiang:
# Parsing operator precedence

Notice how the expressions are built bottom-to-top? The highest-precedence groupings are the smallest. This is easier for humans to see; we can visually parse the expression as a whole. Longer sequences are ... a challenge though.

For a program, it's even harder, especially when we are scanning one token at a time! One way to handle multiple precedence levels is to have multiple functions, one for each level: [f882de85f48e4df27f5525c5b0f1c6cc2c7035ab]

Each function returns a dict representing an expression:

- `'left'`: represents the operand token or expr on the left
- `'oper'`: represents the operator token
- `'right'`: represents the operand token or expr on the right

Since the `'left'` and `'right'` values can be expressions, this gives us a recursive expression tree.

ngjunsiang:
# Parsing functions

The parsing functions call each other, in order of precedence: [253421863c4e98f28c2457ac43d9aae9126ee3ce]

ngjunsiang:
# Helper functions

We add helper functions for parsing tokens and constructing expression dicts: [72896c42292d328c48c9872f87ce29d460400248]

ngjunsiang:
# Recursive descent parsing

`expression()` calls `equality()`, which in turn calls `comparison()`, which calls `addsub()`, which calls `muldiv()`, which calls `value()`.

This ensures that `muldiv()` will never match operators below its precedence in constructing its expression, and so on with each of the parsing functions. This technique is called **recursive descent** parsing: [be32215c251dabb09a9f63a747740b222999d833]

ngjunsiang:
# Testing

Test code:

```
import scanner
import parser

src = '2 * 3 - 1 = 1 / 5 - 4'

tokens = scanner.scan(src)

expr = parser.expression(tokens)

print(expr)
```

Result:

```
{
  'left': {
    'left': {
      'left': {'type': 'integer', 'word': '2', 'value': 2},
      'oper': {'type': 'symbol', 'word': '*', 'value': <function mul at 0x7f3ac0747430>},
      'right': {'type': 'integer', 'word': '3', 'value': 3}
    },
    'oper': {'type': 'symbol', 'word': '-', 'value': <function sub at 0x7f3ac07473a0>},
    'right': {'type': 'integer', 'word': '1', 'value': 1}
  },
  'oper': {'type': 'symbol', 'word': '=', 'value': <function eq at 0x7f3ac0747820>},
  'right': {
    'left': {
      'left': {'type': 'integer', 'word': '1', 'value': 1},
      'oper': {'type': 'symbol', 'word': '/', 'value': <function div at 0x7f3ac07474c0>},
      'right': {'type': 'integer', 'word': '5', 'value': 5}
    },
    'oper': {'type': 'symbol', 'word': '-', 'value': <function sub at 0x7f3ac07473a0>},
    'right': {'type': 'integer', 'word': '4', 'value': 4}
  }
}
```

ngjunsiang:
# Refactoring

- Both `scanner` and `parser` raise `ParseError`: [e34c9a82487c726853d587736fc928ba03509f1e]
- `scanner` needs access to `KEYWORDS` and other constants: [88f0af65ebcaafb6c1b8fc53d205ff893e95a611]
- `scanner` needs access to operators and the classifier to add them to tokens: [7cfffeca68a92f45a0ad7573d274b44e2db5882d]

All have been moved to a new module, [builtin](https://github.com/nyjc-computing/pseudo/blob/2a4e288ad521c82277b38621fdb62935e4024445/builtin.py). Future components needed by multiple modules will also be placed in [builtin](https://github.com/nyjc-computing/pseudo/blob/2a4e288ad521c82277b38621fdb62935e4024445/builtin.py).

PR #8, opened 2022-04-14
03 Evaluation

ngjunsiang:
Let's **evaluate** the expression to retrieve the resulting value. We'll do that with an `evaluate()` function. (I'm tempted to use `eval()` but there is already a built-in Python function with that name.)

At this point it is handy to remember that `expression()` creates an expression tree where each node is either a `token` or `expr`. Furthermore,
- `token` has a `type` key, while `expr` doesn't
- `expr['left']` is an `expr` or `token`
- `expr['right']` is an `expr` or `token`,
- `expr['oper']` is a `token` (with a function mapped to the `value` key).

This would be easier to document with object-oriented programming, but we're focusing on the principle here and not the organisation of code.

### DIFF
--- a/interpreter.py
+++ b/interpreter.py
@@ -1,0 +1,9 @@
+def evaluate(expr):
+    # Evaluating tokens
+    if 'type' in expr:
+        return expr['value']
+    # Evaluating exprs
+    left = evaluate(expr['left'])
+    right = evaluate(expr['right'])
+    oper = expr['oper']['value']
+    return oper(left, right)


ngjunsiang:
https://github.com/nyjc-computing/pseudo/blob/e60de81110c1e9264071db257532a9e1c92e07c4/interpreter.py#L1-L9

Notice that `evaluate()` calls itself to evaluate its `left` and `right` branches, reaching its base case when it encounters a value token which contains a Python value in its `value` key. Since the entire expression tree is made out of dicts, We differentiate tokens from expressions primarily by the value of the `'type'` key.

This would be more elegant with OOP, but I'll leave that for a future commit.

ngjunsiang:
# Testing

```
import scanner
import parser
import interpreter

src = '2 * 3 - 1 = 1 / 5 - 4'

tokens = scanner.scan(src)

expr = parser.expression(tokens)

result = interpreter.evaluate(expr)

print(result)
```

Result:

```False```

PR #10, opened 2022-04-15
05 Interpreting

ngjunsiang:
Let's start changing state by **interpreting** the code.

### DIFF
--- a/builtin.py
+++ b/builtin.py
@@ -2,6 +2,8 @@
 
 class ParseError(Exception): pass
 
+class RuntimeError(Exception): pass
+
 
 
 # Token types
--- a/interpreter.py
+++ b/interpreter.py
@@ -7,3 +7,10 @@ def evaluate(expr):
     right = evaluate(expr['right'])
     oper = expr['oper']['value']
     return oper(left, right)
+
+def execute(stmt):
+    if stmt['rule'] == 'output':
+        for expr in stmt['exprs']:
+            print(str(evaluate(expr)), end='')
+        print('')  # Add line break
+    # Add more if statements for other kinds of statements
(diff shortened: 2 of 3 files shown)


ngjunsiang:
# Executing statements

Let's execute statements as well, starting with the OUTPUT statement we implemented in the previous notebook.

For now, we only have OUTPUT statements to execute. When we have more kinds of statements, they too will go into the `if-else` cascade.

[102ed45028f993714854ee554fd44732e5b052f1]

ngjunsiang:
# Interpreting

Our interpreter runs the following in a loop:
https://github.com/nyjc-computing/pseudo/blob/9dc3aa28dbfd607def076b19591878d0392045bb/interpreter.py#L22-L28

It takes the list of `statement`s from the parser, and `execute()`s them to bring about their effects.

We catch any errors that occur, and raise our own custom `RuntimeError` instead.

ngjunsiang:
# Main loop

The main program loop :

1. passes `src` to the scanner and gets `tokens`,
2. passes `tokens` to the parser and gets `statements`,
3. passes `statements` to the interpreter,
4. exits.

[acc19e752dafd7fead8a5a59b445e9f96c787a5d]

ngjunsiang:
# Exit codes

We also attempt to use proper system exit codes as documented in https://nixdoc.net/man-pages/FreeBSD/man3/sysexits.3.html, but this is not the focus for now.

https://github.com/nyjc-computing/pseudo/blob/acc19e752dafd7fead8a5a59b445e9f96c787a5d/main.py#L20

https://github.com/nyjc-computing/pseudo/blob/acc19e752dafd7fead8a5a59b445e9f96c787a5d/main.py#L25

ngjunsiang:
# Testing

```
if __name__ == "__main__":
    main()
```

Result:

```
Hello, everyone!
```

[39779f87cc40d89072cdb6f009dc060956e929f1]

PR #11, opened 2022-04-15
06a Variables

ngjunsiang:
# Variables

In 9608 pseudocode, variables are declared like this:

    DECLARE Index : INTEGER
    DECLARE Answer : STRING

Values are assigned to the variables as follows:

    Index <- 1
    Answer <- "42"

### DIFF
--- a/interpreter.py
+++ b/interpreter.py
@@ -5,6 +5,8 @@ from builtin import RuntimeError
 def evaluate(expr):
     # Evaluating tokens
     if 'type' in expr:
+        if expr['type'] == 'name':
+            return expr['word']
         return expr['value']
     # Evaluating exprs
     left = evaluate(expr['left'])
@@ -20,6 +22,7 @@ def execute(frame, stmt):
     if stmt['rule'] == 'declare':
         name = evaluate(stmt['name'])
         type_ = evaluate(stmt['type'])
+        breakpoint()
         frame[name] = type_
 
 def interpret(statements):
--- a/interpreter.py
+++ b/interpreter.py
@@ -12,12 +12,15 @@ def evaluate(expr):
     oper = expr['oper']['value']
     return oper(left, right)
 
-def execute(stmt):
+def execute(frame, stmt):
     if stmt['rule'] == 'output':
         for expr in stmt['exprs']:
             print(str(evaluate(expr)), end='')
         print('')  # Add line break
-    # Add more if statements for other kinds of statements
+    if stmt['rule'] == 'declare':
+        name = evaluate(stmt['name'])
+        type_ = evaluate(stmt['type'])
+        frame[name] = type_
 
 def interpret(statements):
     frame = {}
--- a/interpreter.py
+++ b/interpreter.py
@@ -27,3 +27,4 @@ def interpret(statements):
         except RuntimeError:
             print()
             break
+    return frame
--- a/interpreter.py
+++ b/interpreter.py
@@ -20,9 +20,10 @@ def execute(stmt):
     # Add more if statements for other kinds of statements
 
 def interpret(statements):
+    frame = {}
     for stmt in statements:
         try:
-            execute(stmt)
+            execute(frame, stmt)
         except RuntimeError:
             print()
             break
--- a/interpreter.py
+++ b/interpreter.py
@@ -22,7 +22,6 @@ def execute(frame, stmt):
     if stmt['rule'] == 'declare':
         name = evaluate(stmt['name'])
         type_ = evaluate(stmt['type'])
-        breakpoint()
         frame[name] = type_
 
 def interpret(statements):
--- a/main.py
+++ b/main.py
@@ -27,4 +27,13 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    # main()
+    tokens = scanner.scan('DECLARE Index : INTEGER')
+
+    for token in tokens:
+        print(token)
+    
+    statements = parser.parse(tokens)
+    
+    for statement in statements:
+        print(statement)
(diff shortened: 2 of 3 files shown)


ngjunsiang:
# Testing DECLARE statement

Let's see what tokens we get with the above statements.
https://github.com/nyjc-computing/pseudo/blob/1c850b543fee7283eb5c2571519f2e0a5ce280b5/main.py#L31-L39

Result:

```
{'type': 'keyword', 'word': 'DECLARE', 'value': None}
{'type': 'name', 'word': 'Index', 'value': None}
{'type': 'symbol', 'word': ':', 'value': None}
{'type': 'name', 'word': 'INTEGER', 'value': None}
{'type': 'keyword', 'word': '\n', 'value': None}
  File "main.py", line 36, in <module>
    statements = parser.parse(tokens)
  File "/home/runner/pseudo/parser.py", line 130, in parse
    statements += [statement(tokens)]
  File "/home/runner/pseudo/parser.py", line 122, in statement
    raise ParseError(f"Unrecognised token {check(tokens)}")
builtin.ParseError: Unrecognised token {'type': 'keyword', 'word': 'DECLARE', 'value': None}
```

Oops ... our parser can't handle DECLARE statements yet.

ngjunsiang:
# Detecting variable names

We need an expression precedence function for tokenising single names (for now).

Since all of them except `value()` are expecting multiple tokens to form exprs, it seems like we can put it in `value()`:
https://github.com/nyjc-computing/pseudo/blob/6aa26a64f80d7eb82b73092288343a9dd60a2ee5/parser.py#L44-L45

For now, this just returns a `name`-type token without a `value`.

ngjunsiang:
# Handling unexpected tokens

https://github.com/nyjc-computing/pseudo/blob/1a99ea66722454285692ac4b4f2e1c61374f301e/parser.py#L46-L47

As we implement more and more statements and check other kinds of pseudocode statements, we want to know if we encounter tokens whose parsing we have not implemented yet.

ngjunsiang:
# Parsing a DECLARE statement

https://github.com/nyjc-computing/pseudo/blob/e8925fd0994b2ff0d0c863d844c38ad01a71690d/parser.py#L132-L133
We add an `if` case to match DECLARE tokens in `statement()`.

https://github.com/nyjc-computing/pseudo/blob/aaefbff0d4c982e59a0f710ca96c511203be4b50/parser.py#L124-L127
Our parser expects a `value` after DECLARE; this could end up being an integer/string instead of a name, but we'll type-check that later in the interpreter or something.

After that, it consumes a colon (`:`), a token (which we assume to be a type token (i.e. `INTEGER`/`STRING`) but we won't type-check this just yet), and a line break (`\n`).

https://github.com/nyjc-computing/pseudo/blob/aaefbff0d4c982e59a0f710ca96c511203be4b50/parser.py#L128-L132
Like with OUTPUT statements, we wrap the DECLARE in a dict, with the rule `'declare'`.

ngjunsiang:
# The global frame

We have expressions that can be evaluated to yield a value, but this isn't an expression. We want *something* which can remember names and the values they have been assigned. The memory allocated for doing so is usually called the **frame**, so let's stick with that terminology so we don't have to invent our own.

Where should we put our frame?  
`main()` sends the source code string to `scan()` to get a list of `token`s,  
which get sent to `parse()` to get a list of `statement`s,  
which are then sent to `interpret()` to be interpreted (i.e. have their effects carried out or remembered).

Remembering a name and its value should be considered an effect, since we don't actually know what the code means while we are scanning and parsing (we are just building up the structure of the code as a tree). So, instinctively, we put the frame in the interpreter.

Since we are mapping names to values, a dict sounds like a reasonable data structure to use, until we need something more powerful:

[ddf16c06ffeb4a724dcf595f67893a081977b90f]

ngjunsiang:
# Inspecting the frame

https://github.com/nyjc-computing/pseudo/blob/71c89ee68bb9591245ef266e245dad0052a830fe/interpreter.py#L30
This line is a concession for our debugging needs, which we will remove when we are done. We have no way of knowing what happened in the frame otherwise.

ngjunsiang:
# Bugfixes

Error:
```
  File "parser.py", line 34, in value
    if tokens['type'] in ['integer', 'string']:
TypeError: list indices must be integers or slices, not str
```

typo: `tokens` --> `token`

ngjunsiang:
# Executing a DECLARE statement

We can detect names with `value()`, we parse DECLARE statements with `declareStmt()` (inside `statement()`), and finally we can try to execute them. This involves:
1. `evaluate()`ing the `'name'`-type token to get the name (as a string),
2. `evaluate()`ing the `expr` to get the value

https://github.com/nyjc-computing/pseudo/blob/c7240d2ff9f504af549fe47687d5584844736851/interpreter.py#L20-L23

ngjunsiang:
# Testing

https://github.com/nyjc-computing/pseudo/blob/a02270f23ee40e2f3a303cd406b905ebf9624bce/main.py#L10

https://github.com/nyjc-computing/pseudo/blob/a02270f23ee40e2f3a303cd406b905ebf9624bce/main.py#L21-L23

https://github.com/nyjc-computing/pseudo/blob/a02270f23ee40e2f3a303cd406b905ebf9624bce/main.py#L30-L31

Result:

```
{None: None}
```

ngjunsiang:
# Bugfix

https://github.com/nyjc-computing/pseudo/blob/9eaeba9d00f4127071f05fa54fae5bc2dd3eebbc/interpreter.py#L8-L9

Result:

```
{'Index': 'INTEGER'}
```

ngjunsiang:
This will do for now.

## Next steps

1. Assigning values to variables
2. Evaluating variables

PR #12, opened 2022-04-15
06b Assignment

ngjunsiang:
Let's test how our code handles assigning of variables:

https://github.com/nyjc-computing/pseudo/blob/55b9ab7fbb6d0a792d3a164481268017bcb8d7ee/main.py#L10-L13

### DIFF
--- a/builtin.py
+++ b/builtin.py
@@ -4,7 +4,7 @@ class ParseError(Exception): pass
 
 class RuntimeError(Exception): pass
 
-class LogicError(Expcetion): pass
+class LogicError(Exception): pass
 
 
 
--- a/builtin.py
+++ b/builtin.py
@@ -4,6 +4,8 @@ class ParseError(Exception): pass
 
 class RuntimeError(Exception): pass
 
+class LogicError(Expcetion): pass
+
 
 
 # Token types
--- a/interpreter.py
+++ b/interpreter.py
@@ -28,6 +28,13 @@ def execute(frame, stmt):
         value = evaluate(stmt['expr'])
         if name not in frame:
             raise LogicError(f'Undeclared name {repr(name)}')
+        # HACK: type-check values before storing
+        frametype = frame[name]['type']
+        valuetype = type(value)
+        if frametype == 'INTEGER' and valuetype != int:
+            raise LogicError(f'Expected {frametype}, got {valuetype}')
+        elif frametype == 'STRING' and valuetype != str:
+            raise LogicError(f'Expected {frametype}, got {valuetype}')
         frame[name]['value'] = value
 
 def interpret(statements):
--- a/interpreter.py
+++ b/interpreter.py
@@ -14,28 +14,37 @@ def evaluate(expr):
     oper = expr['oper']['value']
     return oper(left, right)
 
+def execOutput(frame, stmt):
+    for expr in stmt['exprs']:
+        print(str(evaluate(expr)), end='')
+    print('')  # Add \n
+
+def execDeclare(frame, stmt):
+    name = evaluate(stmt['name'])
+    type_ = evaluate(stmt['type'])
+    frame[name] = {'type': type_, 'value': None}
+
+def execAssign(frame, stmt):
+    name = evaluate(stmt['name'])
+    value = evaluate(stmt['expr'])
+    if name not in frame:
+        raise LogicError(f'Undeclared name {repr(name)}')
+    # HACK: type-check values before storing
+    frametype = frame[name]['type']
+    valuetype = type(value)
+    if frametype == 'INTEGER' and valuetype != int:
+        raise LogicError(f'Expected {frametype}, got {valuetype}')
+    elif frametype == 'STRING' and valuetype != str:
+        raise LogicError(f'Expected {frametype}, got {valuetype}')
+    frame[name]['value'] = value
+
 def execute(frame, stmt):
     if stmt['rule'] == 'output':
-        for expr in stmt['exprs']:
-            print(str(evaluate(expr)), end='')
-        print('')  # Add \n
+        execOutput(frame, stmt)
     if stmt['rule'] == 'declare':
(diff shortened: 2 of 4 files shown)


ngjunsiang:
Result:

```
Unrecognised token {'type': 'keyword', 'word': '\n', 'value': None}
repl process died unexpectedly: exit status 65
```

Oops, it's choking on the first line break.

https://github.com/nyjc-computing/pseudo/blob/9eaeba9d00f4127071f05fa54fae5bc2dd3eebbc/parser.py#L148-L149

`parse()` is not expecting to encounter line breaks at the start of the code.

ngjunsiang:
# Bugfix

[34e8cfa533a705196ab8ce4984635612992c82f3]

We use a `while` loop to match and discard line breaks at the start of the code.

Result:

```
Unrecognised token {'type': 'name', 'word': 'Index', 'value': None}
repl process died unexpectedly: exit status 65
```

Now we need our parser to recognise assignment statements.

ngjunsiang:
# Assignment statements

These start with a `name` token, followed by `<-`, followed by a value or an expression that evaluates to a value.

https://github.com/nyjc-computing/pseudo/blob/68284e3c15afd5fa3683c3b3914ade6d1a907548/parser.py#L127-L136

ngjunsiang:
# Parsing assignment statements

These do not start with a keyword, but with a `name` token. For now, we assume all statements starting with `name` tokens are supposed to be assignment statements:

https://github.com/nyjc-computing/pseudo/blob/2fe6b464a966880b21fa5b5e209e2771385e0235/parser.py#L143-L144

ngjunsiang:
# Executing assignments

https://github.com/nyjc-computing/pseudo/blob/e4443299e0e2c898e440951ae5d630c4c372f59b/interpreter.py#L26-L29

We execute assignment statements by storing the evaluated value under the evaluated name in the frame.

But this overwrites the type data which was stored by a previous DECLARE statement:
https://github.com/nyjc-computing/pseudo/blob/e4443299e0e2c898e440951ae5d630c4c372f59b/interpreter.py#L22-L25

ngjunsiang:
# Preparing for type-checking

9608 pseudocode does not allowing assigning values to names if the value type does not match the name's declared type. So we should not allow users to do so; let's throw a `LogicError` here instead.

We need a way to store both type and value data. In the interest of simplicity ... let's store them both in the frame, bundled into a dict:

https://github.com/nyjc-computing/pseudo/blob/08945061e4b50cfec4faa5f243b9cd6648e9ce7b/interpreter.py#L22-L25

https://github.com/nyjc-computing/pseudo/blob/08945061e4b50cfec4faa5f243b9cd6648e9ce7b/interpreter.py#L26-L31

ngjunsiang:
# Type-checking

For now, we implement some rudimentary type-checking. Before assigning values, we check that they are of the correct type as previously declared in the DECLARE statement.

https://github.com/nyjc-computing/pseudo/blob/4771bdf791d299d2b2b10767d2e98c5466e7346c/interpreter.py#L31-L37

We will need a more coherent system for enforcing static type consistency, but we don't need that for now.

ngjunsiang:
# Refactor

I can foresee that the `execute()` function is going to get longer and more complicated as we add more statements.

To make the logic clearer, let's use multiple dispatch to break up the code here. Each statement executor gets its own function, named according to the statement it executes and following the pattern `exec*()`.

We call these functions from `execute()` based on the outcome of its token matching:

https://github.com/nyjc-computing/pseudo/blob/e1de6a385c41030f287497c1ecf3e4d501a6e913/interpreter.py#L17-L47

ngjunsiang:
# Bugfix

Let's test the program again:

```
Unrecognised token {'type': 'EOF', 'word': '', 'value': None}
repl process died unexpectedly: exit status 65
```

Hmm, strange ... our EOF token is supposed to be caught by `atEnd()` in `parse()`:
https://github.com/nyjc-computing/pseudo/blob/35bcea54b6bb1ae1dd1429705a8253c684af4f66/parser.py#L150-L157

ngjunsiang:
# Bugtrace

I suspect the extra line breaks at the end of the program are causing the loop to continue even when there are only line breaks left. Let's put in a `print()` to check:
[d851cc458e0577f92adb6e556e2b4404a637571e]

Result:
```
[{'type': 'name', 'word': 'Index', 'value': None}, {'type': 'symbol', 'word': '<-', 'value': None}, {'type': 'integer', 'word': '1', 'value': 1}, {'type': 'keyword', 'word': '\n', 'value': None}, {'type': 'keyword', 'word': '\n', 'value': None}, {'type': 'EOF', 'word': '', 'value': None}]
[{'type': 'keyword', 'word': '\n', 'value': None}, {'type': 'keyword', 'word': '\n', 'value': None}, {'type': 'EOF', 'word': '', 'value': None}]
Unrecognised token {'type': 'EOF', 'word': '', 'value': None}
repl process died unexpectedly: exit status 65
```

Yep, we can see that in the last line before the error, there were two `\n` tokens and an `EOF` token left. so `atEnd()` will not return `True`.

ngjunsiang:
# Bugfix

We repeat the same trick, matching and discarding line break tokens after parsing a statement.
https://github.com/nyjc-computing/pseudo/blob/0b907e803b5fca9764ad670875e5d3e70ee1161c/parser.py#L157-L158

Result:
```
{'Index': {'type': 'INTEGER', 'value': 1}}
```

ngjunsiang:
# LogicError testing

Let's see if our program behaves correctly if we attempt to assign to a variable before declaring it:
https://github.com/nyjc-computing/pseudo/blob/789364516098a691b089985f98a765ee4e6047f5/main.py#L10-L12

Result:
```
  File "interpreter.py", line 31, in execAssign
    raise LogicError(f'Undeclared name {repr(name)}')
builtin.LogicError: Undeclared name 'Index'
```
Yep, we get a `LogicError` as expected.

PR #13, opened 2022-04-15
06c Retrieving variables

ngjunsiang:
# Testing

Let's see how our program handles variables being used in place of values:
https://github.com/nyjc-computing/pseudo/blob/ecc694bbfabc210fd9c85bb336278320f9061780/main.py#L10-L14

Result:
```
Index
{'Index': {'type': 'INTEGER', 'value': 1}}
```

This is because `evaluate()` handles `name` tokens by returning their `word` value:
https://github.com/nyjc-computing/pseudo/blob/ecc694bbfabc210fd9c85bb336278320f9061780/interpreter.py#L5-L9

### DIFF
--- a/builtin.py
+++ b/builtin.py
@@ -60,7 +60,7 @@ def eq(x, y):
     return x == y
 
 def get(frame, name):
-    return frame[name]
+    return frame[name]['value']
 
 
 
--- a/builtin.py
+++ b/builtin.py
@@ -8,33 +8,6 @@ class LogicError(Exception): pass
 
 
 
-# Token types
-
-KEYWORDS = [
-    'DECLARE', 'OUTPUT',
-    'WHILE', 'DO', 'ENDWHILE',
-    'CASE', 'OF', 'OTHERWISE', 'ENDCASE',
-]
-
-TYPES = ['INTEGER', 'STRING']
-
-OPERATORS = {
-    '+': add,
-    '-': sub,
-    '*': mul,
-    '/': div,
-    '<': lt,
-    '<=': lte,
-    '>': gt,
-    '>=': gte,
-    '<>': ne,
-    '=': eq,
-}
-
-SYMBOLS = [':']
-
-
-
 # Operators
 
 def add(x, y):
@@ -69,3 +42,30 @@ def eq(x, y):
 
 def get(frame, name):
     return frame[name]['value']
+
+
+
+# Token types
+
+KEYWORDS = [
+    'DECLARE', 'OUTPUT',
+    'WHILE', 'DO', 'ENDWHILE',
+    'CASE', 'OF', 'OTHERWISE', 'ENDCASE',
+]
+
+TYPES = ['INTEGER', 'STRING']
+
+OPERATORS = {
+    '+': add,
+    '-': sub,
+    '*': mul,
+    '/': div,
+    '<': lt,
+    '<=': lte,
+    '>': gt,
+    '>=': gte,
+    '<>': ne,
+    '=': eq,
+}
+
+SYMBOLS = [':']
--- a/builtin.py
+++ b/builtin.py
@@ -18,10 +18,18 @@ KEYWORDS = [
 
 TYPES = ['INTEGER', 'STRING']
 
-OPERATORS = [
-    '+', '-', '/', '*', '=',
-    '<', '<-', '<=', '>', '>=', '<>',
-]
+OPERATORS = {
+    '+': add,
+    '-': sub,
+    '*': mul,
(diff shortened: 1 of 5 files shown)


ngjunsiang:
# Get expressions

We need a new kind of `expr`: one that says "please look up this word in a frame, and **get** its assigned value". Let's call it a `'get'` expr.

## Handcrafting a get expr

So far we have been treating `expr`s as binary expressions, which involve a `left` expression/token, an `oper` operator, and a `right` expression/token. Let's see if we can continue this pattern so we don't have to invent a new kind of `expr`.

A `get` involves:
- getting a `frame` (which contains mappings of names to values),
- getting a `name`,
- getting the value associated with the `name` from the `frame`

That's promising; `get` lends itself well to being a binary operator. We can try to define it like `get(frame, name)`:
https://github.com/nyjc-computing/pseudo/blob/99b4d015362d2de1463f0ad04cc8650f0e0ca873/builtin.py#L62-L63

ngjunsiang:
# Parsing names

We add a new parser for `name`s, which we call `identifier()` because we are using `name` as a variable name:
https://github.com/nyjc-computing/pseudo/blob/e15fb417c475fe94bc1debeb0bdbbe35fb373177/parser.py#L31-L36

ngjunsiang:
Our declare and assignment statements also need updating so they get a `name` token and not a `get` expr for the first token:
[737a3e4137844f3045ae26dbb9b972c587363f5d]

ngjunsiang:
# Get parsing precedence

Where should we slot this `'get'` expr in recursive descent parsing?

We are likely to be writing code like `i <- i + 1`, and want `i` to be treated as its assigned value rather than a string or keyword, so that means it should have the same precedence as `value()` (where we capture integers, reals, and strings too).
[26a13acbb321495649eb7dcc8581a952053a412b]

ngjunsiang:
# Frame insertion

For now we will insert the `frame` in the interpreter, when we `evaluate()` a `get` expr:
https://github.com/nyjc-computing/pseudo/blob/7835fc47de7d1ee83a6326a8d515bcdcf03778b3/interpreter.py#L11-L18

ngjunsiang:
Don't forget to update the `evaluate()` calls to add the frame argument:
[bf07c087c1d2fa59222128c21dc54559f096819e]

ngjunsiang:
# Bug

It's not working:

```
  File "main.py", line 26, in main
    frame = interpreter.interpret(statements)
  File "interpreter.py", line 57, in interpret
    execute(frame, stmt)
  File "interpreter.py", line 47, in execute
    execOutput(frame, stmt)
  File "interpreter.py", line 23, in execOutput
    print(str(evaluate(expr)), end='')
  File "interpreter.py", line 19, in evaluate
    return oper(left, right)
  File "builtin.py", line 63, in get
    return frame[name]
TypeError: 'NoneType' object is not subscriptable
```

Suspect the line in `evaluate()` that handles `get` operator:
https://github.com/nyjc-computing/pseudo/blob/bf07c087c1d2fa59222128c21dc54559f096819e/interpreter.py#L13

With a breakpoint there:

```
> interpreter.py(15)evaluate()
-> left = frame  # <-- expr['left'] is None
(Pdb) expr
{'left': None, 'oper': {'type': 'symbol', 'word': '', 'value': <function get at 0x7f66262c99d0>}, 'right': {'type': 'name', 'word': 'Index', 'value': None}}
(Pdb) 
```

`'left'` didn't receive a frame for a `get` expr.

ngjunsiang:
# Fix

Add `frame` into the `evaluate()` call in `execOutput()`:
[f79b55564fd540674ce0033ba6069234cb375123]

Result:

```
{'type': 'INTEGER', 'value': 1}
{'Index': {'type': 'INTEGER', 'value': 1}}
```

ngjunsiang:
Result:

```
1
{'Index': {'type': 'INTEGER', 'value': 1}}
```

Now our program can declare variables, assign values to variables, and subsequently retrieve (evaluate) them.

ngjunsiang:
# Refactor

We have an `OPERATORS` constant containing a list of operator symbols:
https://github.com/nyjc-computing/pseudo/blob/f609d3ab0dc29628c060820baf4776ac26bad510/builtin.py#L21-L24

And we have an `operators` dict mapping operator symbols to operator functions:
https://github.com/nyjc-computing/pseudo/blob/f609d3ab0dc29628c060820baf4776ac26bad510/builtin.py#L69-L80

We don't need both: [58e5ded13b83ea2471e905d554b2c29fdd3007d7]

PR #17, opened 2022-04-17
09 Conditionals

ngjunsiang:
With frames and static typing in place, we can move on to implementing the rest of the statements. We start with the basics: conditionals.

We start by detecting CASE and IF statements in `statement()`:
https://github.com/nyjc-computing/pseudo/blob/9bb4ad42c9da5caab42934212b6bc3f80ddf68fb/parser.py#L159-L162

### DIFF
--- a/interpreter.py
+++ b/interpreter.py
@@ -32,6 +32,12 @@ def execAssign(frame, stmt):
     value = evaluate(stmt['expr'], frame)
     frame[name]['value'] = value
 
+def execCase(frame, stmt):
+    pass
+
+def execIf(frame, stmt):
+    pass
+
 def execute(frame, stmt):
     if stmt['rule'] == 'output':
         execOutput(frame, stmt)
@@ -39,6 +45,10 @@ def execute(frame, stmt):
         execDeclare(frame, stmt)
     if stmt['rule'] == 'assign':
         execAssign(frame, stmt)
+    if stmt['rule'] == 'case':
+        execCase(frame, stmt)
+    if stmt['rule'] == 'if':
+        execIf(frame, stmt)
 
 def interpret(statements, frame=None):
     if frame is None:
--- a/interpreter.py
+++ b/interpreter.py
@@ -33,7 +33,11 @@ def execAssign(frame, stmt):
     frame[name]['value'] = value
 
 def execCase(frame, stmt):
-    pass
+    cond = evaluate(stmt['cond'], frame)
+    if cond in stmt['stmts']:
+        execute(frame, stmt['stmts'][cond])
+    elif stmt['fallback']:
+        execute(frame, stmt['fallback']
 
 def execIf(frame, stmt):
     pass
--- a/interpreter.py
+++ b/interpreter.py
@@ -40,7 +40,12 @@ def execCase(frame, stmt):
         execute(frame, stmt['fallback']
 
 def execIf(frame, stmt):
-    pass
+    if evaluate(stmt['cond'], frame):
+        for substmt in stmt['stmts'][True]:
+            execute(frame, substmt)
+    elif stmt['fallback']:
+        for substmt in stmt['fallback']:
+            execute(frame, substmt)
 
 def execute(frame, stmt):
     if stmt['rule'] == 'output':
--- a/interpreter.py
+++ b/interpreter.py
@@ -37,7 +37,7 @@ def execCase(frame, stmt):
     if cond in stmt['stmts']:
         execute(frame, stmt['stmts'][cond])
     elif stmt['fallback']:
-        execute(frame, stmt['fallback']
+        execute(frame, stmt['fallback'])
 
 def execIf(frame, stmt):
     if evaluate(stmt['cond'], frame):
--- a/main.py
+++ b/main.py
@@ -10,8 +10,16 @@ import interpreter
 
 src = '''
 DECLARE Index : INTEGER
-Index <- "5"
-OUTPUT Index
+Index <- 5
+CASE OF Index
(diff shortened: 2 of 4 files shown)


ngjunsiang:
Before we implement the parsing functions `caseStmt` and `ifStmt`, let's think about what kind of `stmt` we should return.

## CASE

CASE statements are written as follows:

```
CASE OF <identifier>
    <value 1> : <statement>
    <value 2> : <statement>
    ...
    OTHERWISE <statement>
ENDCASE
```

## IF

IF statements are written as follows:

```
IF <condition>
  THEN
    <statements>
  ELSE
    <statements>
ENDIF
```

I'd like to try to minimise the number of different `stmt` types we have to juggle. Can we represent both kinds of statement with one `stmt` format? I think we can.

What they share in common:
- They begin by checking a condition: an identifier for CASE, and a condition for IF
- They execute one of multiple outcomes depending on the result of the condition
- They have an optional fallback outcome (OTHERWISE/ELSE)
- They end with an END* keyword.

ngjunsiang:
# Representing conditionals

Let's represent both with the following `stmt` format:

```
{
    'rule': 'case' | 'if',
    'cond': <name> | <expr>,
    'stmts': {<value>: <statements>},
    'fallback': <statements>,
}
```

ngjunsiang:
# Parsing conditionals

Our two parsing functions:
https://github.com/nyjc-computing/pseudo/blob/4085e5e6d5fc2642b40f46f48eefa4bce292d6a7/parser.py#L149-L170

https://github.com/nyjc-computing/pseudo/blob/4085e5e6d5fc2642b40f46f48eefa4bce292d6a7/parser.py#L172-L198

ngjunsiang:
# Executing conditionals

Execution is much more straightforward. We evaluate `cond` and check for a matching key in `stmts`. If there are no matching keys, execute the fallback if there is one.

https://github.com/nyjc-computing/pseudo/blob/7dd99c2bb2febfb4c12cbf6093a1aff090fc4969/interpreter.py#L35-L40

CASE options only involve a single statement, while IF options can contain multiple statements. So we execute the latter in a finite loop.

https://github.com/nyjc-computing/pseudo/blob/268d1a7b579a157870c34a4c0a476d47438691fb/interpreter.py#L42-L48

ngjunsiang:
# Resolving conditionals

We can't just write an executor anymore, we need verifiers for the resolver as well. Otherwise we won't be able to retrieve the value of variables from the frame.

https://github.com/nyjc-computing/pseudo/blob/conditionals/resolver.py#L52-L65

ngjunsiang:
# Type checking for conditionals

The IF conditional check must return a boolean, so let's check for that:

https://github.com/nyjc-computing/pseudo/blob/d4b591e933e8f4b478d1b1ccd589163a068f3c69/resolver.py#L60-L62

ngjunsiang:
# Testing and bugfixing

Lots of little typos and fixes, summarised above.

Test code:

```
DECLARE Index : INTEGER
Index <- 5
CASE OF Index
    5: OUTPUT "Five"
    4: OUTPUT "Four"
    3: OUTPUT "Three"
    2: OUTPUT "Two"
    1: OUTPUT "One"
    0: OUTPUT "I don't know!"
    OTHERWISE OUTPUT "I can't count beyond five"
ENDCASE
```

Result:
```
Five
```

Frame:
```
{'Index': {'type': 'INTEGER', 'value': 5}}
```

ngjunsiang:
# Testing & Bugfix

Code:
```
DECLARE Index : INTEGER
Index <- 5
IF Index > 5
  THEN
    OUTPUT "I can't count beyond five"
    Index <- 3
  ELSE
    OUTPUT "A small number"
    Index <- 0
ENDIF
```

This is a tricky bug. Here's the IF statement:
```
{'rule': 'if',
 'cond': {'left': {'left': None,
                   'oper': {'type': 'symbol', 'word': '', 'value': <function get at 0x7f8dea056940>},
                   'right': {'type': 'name', 'word': 'Index', 'value': None}},
          'oper': {'type': 'symbol', 'word': '>', 'value': <function gt at 0x7f8dea056700>},
          'right': {'type': 'integer', 'word': '5', 'value': 5}},
 'stmts': {True: [{'rule': 'output', [{'type': 'string', 'word': '"I can\'t count beyond five"', 'value': "I can't count beyond five"}]},
                        {'rule': 'assign', 'name': {'type': 'name', 'word': 'Index', 'value': None}, 'expr': {'type': 'integer', 'word': '3', 'value': 3}}]},
 'fallback': [{'rule': 'output', 'exprs': [{'type': 'string', 'word': '"A small number"', 'value': 'A small number'}]},
              {'rule': 'assign', 'name': {'type': 'name', 'word': 'Index', 'value': None}, 'expr': {'type': 'integer', 'word': '0', 'value': 0}}]}
```

When this statement encounters the resolver and is passed to `resolve()`:
https://github.com/nyjc-computing/pseudo/blob/3b9d3d08bd197a3492b9a839b5eb7a844be51e61/resolver.py#L18-L26

The IF condition encounters line 25 and is returned ... without resolving the expr's `'left'` and `'right'`!

The fix: [25602e5d9bbee1404fd47ef50d8ef39195ec330b]

ngjunsiang:
# Syntax

`resolve()` takes an expr first, followed by frame; meanwhile `verify()` and `execute()` both take the frame as the first argument.

The argument list for `resolve()` has been edited for consistency: [029fbbe6a68c7ed8ef2eb928b50a277c86f6f6b0]

ngjunsiang:
We finally have conditional handling in our pseudocode interpreter!

PR #18, opened 2022-04-18
10 Loops

ngjunsiang:
Now that we have conditional handling, we can finally work on loops.

We start by detecting WHILE, REPEAT, and FOR statements in `statement()`:
https://github.com/nyjc-computing/pseudo/blob/cabcf32df001bf023792673c8cc2344829990cab/parser.py#L218-L223

### DIFF
--- a/interpreter.py
+++ b/interpreter.py
@@ -48,10 +48,18 @@ def execIf(frame, stmt):
             execute(frame, substmt)
 
 def execWhile(frame, stmt):
-    pass
+    if stmt['init']:
+        execute(frame, stmt['init'])
+    while evaluate(stmt['cond'], frame) is True:
+        for loopstmt in stmt['stmts']:
+            execute(frame, loopstmt)
 
 def execRepeat(frame, stmt):
-    pass
+    for loopstmt in stmt['stmts']:
+        execute(frame, loopstmt)
+    while evaluate(stmt['cond'], frame) is True:
+        for loopstmt in stmt['stmts']:
+            execute(frame, loopstmt)
 
 def execute(frame, stmt):
     if stmt['rule'] == 'output':
--- a/interpreter.py
+++ b/interpreter.py
@@ -57,7 +57,7 @@ def execWhile(frame, stmt):
 def execRepeat(frame, stmt):
     for loopstmt in stmt['stmts']:
         execute(frame, loopstmt)
-    while evaluate(stmt['cond'], frame) is True:
+    while evaluate(stmt['cond'], frame) is False:
         for loopstmt in stmt['stmts']:
             execute(frame, loopstmt)
 
--- a/interpreter.py
+++ b/interpreter.py
@@ -47,6 +47,12 @@ def execIf(frame, stmt):
         for substmt in stmt['fallback']:
             execute(frame, substmt)
 
+def execWhile(frame, stmt):
+    pass
+
+def execRepeat(frame, stmt):
+    pass
+
 def execute(frame, stmt):
     if stmt['rule'] == 'output':
         execOutput(frame, stmt)
@@ -58,6 +64,10 @@ def execute(frame, stmt):
         execCase(frame, stmt)
     if stmt['rule'] == 'if':
         execIf(frame, stmt)
+    if stmt['rule'] == 'while':
+        execWhile(frame, stmt)
+    if stmt['rule'] == 'repeat':
+        execRepeat(frame, stmt)
 
 def interpret(statements, frame=None):
     if frame is None:
--- a/main.py
+++ b/main.py
@@ -28,7 +28,6 @@ ENDWHILE
 # DECLARE i : INTEGER
 # FOR i <- 1 TO 10
 #     OUTPUT i
-#     i <- i + 1
 # ENDFOR
 # '''
 
--- a/main.py
+++ b/main.py
@@ -9,18 +9,28 @@ import interpreter
 
 
 src = '''
-DECLARE Index : INTEGER
-Index <- 5
-CASE OF Index
-    5: OUTPUT "Five"
-    4: OUTPUT "Four"
-    3: OUTPUT "Three"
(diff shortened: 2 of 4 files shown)


ngjunsiang:
Before we implement the parsing functions `whileStmt`, `repeatStmt`, and `forStmt`, let's think about what kind of `stmt` we should return.

## WHILE

WHILE statements are written as follows:
```
WHILE <condition> DO
    <statements>
ENDWHILE
```

## REPEAT

REPEAT statements are written as follows:
```
REPEAT
    <Statements>
UNTIL <condition>
```
## FOR

FOR statements are written as follows:
```
FOR <identifier> <- <value1> TO <value2>
    <statements>
ENDFOR
```
or
```
FOR <identifier> <- <value1> TO <value2> STEP <increment>
<statements>
ENDFOR
```

What they share in common:
- They have a condition that is evaluated at the start/end of each iteration
- They have a list of statements to be executed while a particular condition evaluates to `TRUE`

What they do differently:
- WHILE evaluates the condition first, then executes the statements
- REPEAT executes the statements first, then evaluates the condition
- FOR constructs a range iterator, and checks that the counter is not greater than the range upper limit

ngjunsiang:
# Representing loops

Let's represent all three with the following `stmt` format:

```
{
    'rule': 'while' | | 'repeat' | 'for',
    'cond': <name> | <expr>,
    'stmts': [<statements>],
```

ngjunsiang:
# Parsing WHILE and REPEAT

WHILE statement:
https://github.com/nyjc-computing/pseudo/blob/54d18c65e55729252cea68175e77daaa2c0ee1ac/parser.py#L200-L215

REPEAT statement:
https://github.com/nyjc-computing/pseudo/blob/54d18c65e55729252cea68175e77daaa2c0ee1ac/parser.py#L217-L231

These are relatively straightforward, if a little tedious.

ngjunsiang:
# Desugaring FOR

Instead of making a new type of `stmt` for FOR statements, we can reuse the existing structure. Although a FOR loop can be rewritten into a WHILE loop, the FOR loop is generally easier to read. Such ease-of-use features in a programming language are often called *syntactic sugar* because they make a programming language "sweeter"/easier to use.

When we parse FOR statements, instead of adding more complexity to our interpreter by making a new type of `stmt`, we can instead create them as existing `'while'` stmts, in a process known as **desugaring**.

A FOR loop like this:
```
FOR <identifier> <- <value1> TO <value2> STEP <increment>
    <statements>
ENDFOR
```

gets desugared to:

```
<identifier> <- <value1>
WHILE <identifier> <= <value2> DO
    <statements>
    <identifier> <- <identifier> + <increment>
ENDWHILE
```

ngjunsiang:
# Parsing FOR loops

We capture parts of the FOR loop as follows:
https://github.com/nyjc-computing/pseudo/blob/7fb0f23fbdd122182f397846fac64b388270054b/parser.py#L233-L251

We generate the `init` statement, reusing as much of our parsing machinery as we can:
https://github.com/nyjc-computing/pseudo/blob/7fb0f23fbdd122182f397846fac64b388270054b/parser.py#L252-L258

Likewise for the loop `cond`:
https://github.com/nyjc-computing/pseudo/blob/7fb0f23fbdd122182f397846fac64b388270054b/parser.py#L259-L265

And then the increment at the end of each iteration of statements:
https://github.com/nyjc-computing/pseudo/blob/7fb0f23fbdd122182f397846fac64b388270054b/parser.py#L266-L274

Finally we construct the FOR loop as a `'while'` `stmt`:
https://github.com/nyjc-computing/pseudo/blob/7fb0f23fbdd122182f397846fac64b388270054b/parser.py#L275-L280

This is ugly and complicated, because we are constructing tokens in a module that shouldn't have to know how that is done. We will attempt to clean this up in a future refactoring.

ngjunsiang:
# Testing

We add some test pseudocode to check our parser:
https://github.com/nyjc-computing/pseudo/blob/0207e4870bd3e10950fdf05aa9b3e7b92bd6562c/main.py#L11-L33

And fix a few bugs in our loop parsers:
[0207e4870bd3e10950fdf05aa9b3e7b92bd6562c]

ngjunsiang:
# Resolving loop statements

In a slight twist from the previous chapter/pull request, we do resolving first before interpreting.

Our decision to desugar and use a common `stmt` structure bears fruit almost immediately: we only need one verifier for all three types of statements:
https://github.com/nyjc-computing/pseudo/blob/fca150c08f28bfff3ce4c21d0a83879b32bff42c/resolver.py#L87-L88

No `'for'` statement handling needed, since a FOR loop is desugared to a `'while'` statement.

Recall that the resolver only needs to resolve each expression/statement once, to ensure that types match up and to insert a frame for each `get` expr.

And then we follow the code examples in the rest of `resolver.py` to `verify()` the `stmt['init']` (if there's one), `resolve()` the `stmt['cond']`, and again `verify()` each statement in `stmt['stmts']`:
https://github.com/nyjc-computing/pseudo/blob/396a9e1960fcd8f5357701710346b814db1cef8c/resolver.py#L73-L79

ngjunsiang:
# Interpreting loop statements

The execution is much more straightforward than the parsing:

## WHILE statements
https://github.com/nyjc-computing/pseudo/blob/24e4b0438fd844bb53155a37c34f221f11d0f701/interpreter.py#L50-L55

## REPEAT statements
https://github.com/nyjc-computing/pseudo/blob/24e4b0438fd844bb53155a37c34f221f11d0f701/interpreter.py#L57-L62

We are not expecting any init for REPEAT statements, so we can skip that. And instead of `evaluate()`ing `cond` for `True`, we keep running `stmts` *until* `cond` `evaluate()`s to `False`.

ngjunsiang:
And now we have working loops! We'll catch the detailed bugs later, at least our simple test code runs as expected.

PR #19, opened 2022-04-20
11 Input

ngjunsiang:
A short chapter before we bite off the next big chunk.

We have OUTPUT statements to display text, now we implement INPUT to take in text.

Parsing condition:
https://github.com/nyjc-computing/pseudo/blob/7fa5f74e5efc3fe5937b6288d16b23d745cac42b/parser.py#L289-L290

And parser implementation:
https://github.com/nyjc-computing/pseudo/blob/150cfaa75c5ad2d4b712a1d3001ef51f570b6635/parser.py#L125-L132

### DIFF
--- a/interpreter.py
+++ b/interpreter.py
@@ -22,6 +22,10 @@ def execOutput(frame, stmt):
         print(str(evaluate(expr, frame)), end='')
     print('')  # Add \n
 
+def execInput(frame, stmt):
+    name = evaluate(stmt['name'], frame)
+    frame[name]['value'] = input()
+
 def execDeclare(frame, stmt):
     name = evaluate(stmt['name'], frame)
     type_ = evaluate(stmt['type'], frame)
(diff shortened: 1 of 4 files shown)


ngjunsiang:
# Resolving

An INPUT statement looks like:

```
INPUT Variable
```

This assigns the string entered by the user to `Variable`. Since there are not `get` expressions used here, we don't really need to do any frame insertion.

However, at minimum we do need to check that `Variable` has been declared:
https://github.com/nyjc-computing/pseudo/blob/f6a00fe1c218cc9e53d92ea76f6a6cc874d348f2/resolver.py#L43-L46

ngjunsiang:
# Interpreting INPUT statements

Finally, the implementation. Since this takes in user input and assigns it to the given variable, we draw our inspiration for interpreting from `assignStmt()`:
https://github.com/nyjc-computing/pseudo/blob/57340bc80db02e796d7a66a2dcbb28ccc8d46e55/interpreter.py#L25-L27

ngjunsiang:
# Testing

After fixing a missing argument, our test code successfully runs:

```
DECLARE Result : STRING
OUTPUT "Enter some text: "
INPUT Result
OUTPUT "You entered: ", Result
```

Result:
```
Enter some text: 
Hi
You entered: Hi
```

PR #20, opened 2022-04-20
12a Procedures

ngjunsiang:
With branches and loops in place, let's move on to procedures, then functions.

We will have to add some features to allow procedures and functions to have their own frames for locally declared variables (if any). But not here, not yet. For now, our procedures will use the global frame.

A procedure with no parameters is defined as follows:
```
PROCEDURE <identifier>
 <statements>
ENDPROCEDURE
```

A procedure with parameters is defined as follows:
```
PROCEDURE <identifier>(<param1>:<datatype>,<param2>:<datatype>...)
 <statements>
ENDPROCEDURE
```

### DIFF
--- a/interpreter.py
+++ b/interpreter.py
@@ -65,6 +65,14 @@ def execRepeat(frame, stmt):
         for loopstmt in stmt['stmts']:
             execute(frame, loopstmt)
 
+def execProcedure(frame, stmt):
+    name = evaluate(stmt['name'], frame)
+    frame[name] = {
+        'type': 'procedure',
+        'args': stmt['args'],
+        'stmts': stmt['stmts'],
+    }
+
 def execute(frame, stmt):
     if stmt['rule'] == 'output':
         execOutput(frame, stmt)
@@ -82,6 +90,8 @@ def execute(frame, stmt):
         execWhile(frame, stmt)
     if stmt['rule'] == 'repeat':
         execRepeat(frame, stmt)
+    if stmt['rule'] == 'procedure':
+        execProcedure(frame, stmt)
 
 def interpret(statements, frame=None):
     if frame is None:
--- a/main.py
+++ b/main.py
@@ -9,10 +9,10 @@ import interpreter
 
 
 src = '''
-DECLARE Result : STRING
-OUTPUT "Enter some text: "
-INPUT Result
-OUTPUT "You entered: ", Result
+DECLARE Person : STRING
+PROCEDURE SayHi(Person : STRING)
+    OUTPUT "Hi, ", Person, "!"
+ENDPROCEDURE
 '''
 
 
--- a/parser.py
+++ b/parser.py
@@ -293,15 +293,15 @@ def procedureStmt(tokens):
     name = identifier(tokens)
     args = {}
     if match(tokens, '('):
-        name = identifier(tokens)
+        var = identifier(tokens)
         expectElseError(tokens, ':')
         typetoken = consume(tokens)
-        args[name] = {'type': typetoken, 'value': None}
+        args[var] = {'type': typetoken, 'value': None}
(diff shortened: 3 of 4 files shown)


ngjunsiang:
# Parsing PROCEDURE statements

We match for the PROCEDURE keyword:
https://github.com/nyjc-computing/pseudo/blob/256476cdff7a680a934b2f7e77245c517ce16f06/parser.py#L312-L313

And implement it as follows:
https://github.com/nyjc-computing/pseudo/blob/ea894883e1d0dad20319865a7384c113fa68200d/parser.py#L292-L318

`if-while` is an ugly way to parse optional arguments, but we can figure out more elegant ways later. Let's just make it work first.

At first glance, procedures look very similar to functions. Couldn't we just use the same `stmt` structure for both procedures and functions?

Yes, kinda ... but we can only call procedures with the CALL keyword (i.e. calling a procedure is a statement in itself), while functions can be called anywhere in place of an expression.

This means we need a way to differentiate functions and procedures, such that function calls are `evaluate()`d while procedure CALLs are `execute()`d. For now we write them as two separate statements, and later on we can refactor.

ngjunsiang:
# Resolving PROCEDURE statements

This one is easy:
https://github.com/nyjc-computing/pseudo/blob/6621c777162d6e339336a47b20748c25121e188b/resolver.py#L104-L105

https://github.com/nyjc-computing/pseudo/blob/6621c777162d6e339336a47b20748c25121e188b/resolver.py#L86-L88

Procedures have no return type that need matching or storing, and the name and argument declarations have no `get` expressions. That leaves only the `stmts` that need verifying.

ngjunsiang:
# Executing PROCEDURE declarations

Declaring a procedure is the easy part:
https://github.com/nyjc-computing/pseudo/blob/d5a8920b8afd667203d9d21f11b3383a9116391c/interpreter.py#L93-L94

https://github.com/nyjc-computing/pseudo/blob/d5a8920b8afd667203d9d21f11b3383a9116391c/interpreter.py#L68-L74

We just store everything the procedure needs into the frame for now. Executing it is going to be another story ...

And meanwhile we catch a typo bug:
[09d2ecb27e33f710f437950af6fa35ddd3487481]

ngjunsiang:
# Testing

```
PROCEDURE SayHi
    OUTPUT "Hi!"
ENDPROCEDURE
```

Result (frame):
```
{'SayHi': {'type': 'procedure', 'args': {}, 'stmts': [{'rule': 'output', 'exprs': [{'type': 'string', 'word': '"Hi!"', 'value': 'Hi!'}]}]}}
```

```
DECLARE Person : STRING
PROCEDURE SayHi(Person : STRING)
    OUTPUT "Hi, ", Person, "!"
ENDPROCEDURE
```

Result:
```
  File "parser.py", line 299, in procedureStmt
    args[var] = {'type': typetoken, 'value': None}
TypeError: unhashable type: 'dict'
```

Oops. At this point, `var` is a token and not yet a name. I'll have to extract `'word'` from the token first. Fix: [15113393535903855e1fb2b9624e70b719b4445c]

Result (frame):
```
{'Person': {'type': 'STRING', 'value': None}, 'SayHi': {'type': 'procedure', 'args': {'Person': {'type': {'type': 'name', 'word': 'STRING', 'value': None}, 'value': None}}, 'stmts': [{'rule': 'output', 'exprs': [{'type': 'string', 'word': '"Hi, "', 'value': 'Hi, '}, {'left': {...}, 'oper': {'type': 'symbol', 'word': '', 'value': <function get at 0x7f1d6607c940>}, 'right': {'type': 'name', 'word': 'Person', 'value': None}}, {'type': 'string', 'word': '"!"', 'value': '!'}]}]}}
```

PR #24, opened 2022-04-21
12c Passing by reference

ngjunsiang:
9608 pseudocode recognises the notion of passing-by-value vs passing-by-reference.

Procedures are able to access all variables in the global frame. But reassigning variables is another matter.
- Passing-by-reference means that any reassignments to a variable in the local frame affect the global frame as well.
- Passing-by-value means that reassignments to a variable in the local frame do not affect the global frame. These reassignments are therefore lost when the procedure exits and the local frame is no longer used (since 9608 pseudocode does not support closures).

### DIFF
--- a/interpreter.py
+++ b/interpreter.py
@@ -67,14 +67,18 @@ def execProcedure(frame, stmt):
     pass
 
 def execCall(frame, stmt):
+    # Set up local frame
+    local = {}
+    for var, typevalue in frame.items():
+        local[var] = typevalue.copy()
     # Get procedure from frame
     proc = evaluate(frame, stmt['name'])
-    # Assign args into frame with param names
+    # Assign args into local with param names
     args, params = stmt['args'], proc['params']
     for arg, var, param in zip(args, params.keys(), params.values()):
-        frame[var]['value'] = evaluate(frame, arg)
+        local[var]['value'] = evaluate(frame, arg)
     for callstmt in proc['stmts']:
-        execute(frame, callstmt)
+        execute(local, callstmt)
 
 def execute(frame, stmt):
     if stmt['rule'] == 'output':
--- a/interpreter.py
+++ b/interpreter.py
@@ -71,6 +71,7 @@ def execCall(frame, stmt):
     #     'type': 'procedure',
     #     'value': {
     #         'frame': local,
+    #         'passby': stmt['passby']['word'],
     #         'params': stmt['params'],
     #         'stmts': stmt['stmts'],
     #     }
@@ -78,11 +79,12 @@ def execCall(frame, stmt):
     # Get procedure from frame
     proc = evaluate(frame, stmt['name'])
     # Set up local frame
-    local = proc['value']['frame'].copy()
+    local = {name: var for name, var in proc['frame'].items()}
     # Assign args into local with param names
-    args, params = stmt['args'], proc['value']['params']
+    args, params = stmt['args'], proc['params']
     for arg, param in zip(args, params):
-        local[param['name']]['value'] = evaluate(frame, arg)
+        name = evaluate(local, param['name'])
+        local[name]['value'] = evaluate(frame, arg)
     for callstmt in proc['stmts']:
         execute(local, callstmt)
 
--- a/interpreter.py
+++ b/interpreter.py
@@ -67,16 +67,22 @@ def execProcedure(frame, stmt):
     pass
 
 def execCall(frame, stmt):
-    # Set up local frame
-    local = {}
-    for var, typevalue in frame.items():
-        local[var] = typevalue.copy()
+    # frame[name] = {
+    #     'type': 'procedure',
+    #     'value': {
+    #         'frame': local,
+    #         'params': stmt['params'],
+    #         'stmts': stmt['stmts'],
+    #     }
+    # }
     # Get procedure from frame
     proc = evaluate(frame, stmt['name'])
+    # Set up local frame
+    local = proc['value']['frame'].copy()
     # Assign args into local with param names
-    args, params = stmt['args'], proc['params']
-    for arg, var, param in zip(args, params.keys(), params.values()):
-        local[var]['value'] = evaluate(frame, arg)
+    args, params = stmt['args'], proc['value']['params']
+    for arg, param in zip(args, params):
+        local[param['name']]['value'] = evaluate(frame, arg)
     for callstmt in proc['stmts']:
         execute(local, callstmt)
 
--- a/interpreter.py
+++ b/interpreter.py
@@ -71,15 +71,19 @@ def execCall(frame, stmt):
     #     'type': 'procedure',
     #     'value': {
     #         'frame': local,
-    #         'passby': stmt['passby']['word'],
+    #         'passby': str,
     #         'params': stmt['params'],
     #         'stmts': stmt['stmts'],
     #     }
     # }
     # Get procedure from frame
     proc = evaluate(frame, stmt['name'])
-    # Set up local frame
-    local = {name: var for name, var in proc['frame'].items()}
+    # Note: for BYREF variables, the procedure's frame
+    # may still contain values from previous calls.
+    # Do not evaluate any get exprs for local frame
+    # before assigning local variables from args.
+    local = proc['frame']
+    
     # Assign args into local with param names
     args, params = stmt['args'], proc['params']
     for arg, param in zip(args, params):
--- a/main.py
+++ b/main.py
@@ -13,8 +13,10 @@ src = '''
 DECLARE Person : STRING
 PROCEDURE SayHi(BYREF Person : STRING)
     OUTPUT "Hi, ", Person, "!"
+    Person <- "Mary"
 ENDPROCEDURE
 CALL SayHi("John")
+OUTPUT Person
 '''
 
 
--- a/main.py
+++ b/main.py
@@ -11,7 +11,7 @@ import interpreter
 
 src = '''
 DECLARE Person : STRING
-PROCEDURE SayHi(Person : STRING)
+PROCEDURE SayHi(BYREF Person : STRING)
     OUTPUT "Hi, ", Person, "!"
 ENDPROCEDURE
 CALL SayHi("John")
--- a/parser.py
+++ b/parser.py
@@ -133,15 +133,25 @@ def inputStmt(tokens):
     }
     return stmt
 
-def declareStmt(tokens):
+def declare(tokens):
     name = identifier(tokens)
     expectElseError(tokens, ':')
     typetoken = consume(tokens)
+    if typetoken['word'] not in TYPES:
+        raise ParseError(f"Invalid type {typetoken['word']}")
+    var = {
+        'name': name,
+        'type': typetoken,
+    }
+    return var
+    
+def declareStmt(tokens):
+    var = declare(tokens)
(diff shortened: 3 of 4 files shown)


ngjunsiang:
# Using a local frame for CALL execution

To use a local frame, we are going to need to create one when we execute the call:
[c2319a7f5477a4d8eadbe40e504d8334be51b6a2]

https://github.com/nyjc-computing/pseudo/blob/c2319a7f5477a4d8eadbe40e504d8334be51b6a2/interpreter.py#L69-L73
First, we create a `local` frame (just like the global one). We make a copy of each variable already in the global frame (this is pass-by-value).

https://github.com/nyjc-computing/pseudo/blob/c2319a7f5477a4d8eadbe40e504d8334be51b6a2/interpreter.py#L76-L79
Now we assign the `args` to the appropriate named `vars`, according to `params` order.  Notice that when we evaluate the `arg`, we still do so with the global frame, since that is where the function is called and the args are passed.

https://github.com/nyjc-computing/pseudo/blob/c2319a7f5477a4d8eadbe40e504d8334be51b6a2/interpreter.py#L81
Finally, we execute the CALL statement with the `local` frame.

ngjunsiang:
# Designating pass-by-value or pass-by-reference

While 9608 pseudocode assumes pass-by-value as the default mode of passing arguments, it also supports passing-by-reference:

> To specify whether a parameter is passed by value or by reference, the keywords `BYVALUE` and `BYREF`
precede the parameter in the definition of the procedure. If there are several parameters, they should all be
passed by the same method and the `BYVALUE` or `BYREF` keyword need not be repeated.
> **Example – passing parameters by reference**
> ```
> PROCEDURE SWAP(BYREF X : INTEGER, Y : INTEGER)
>     Temp <- X
>     X <- Y
>     Y <- Temp
> ENDPROCEDURE 
> ```
> If the method for passing parameters is not specified, passing by value is assumed.

ngjunsiang:
# Parsing BYVALUE and BYREF

We start with recognising these two new keywords in the parser. We initialise `passby` with a default `BYVALUE` token, then override it if the parser matches an appropriate token after the opening `(`:
https://github.com/nyjc-computing/pseudo/blob/7e828efb545e2742fc05dbd5e17fb168123d5953/parser.py#L297-L300

Our `stmt` dict now gains a new key:
https://github.com/nyjc-computing/pseudo/blob/7e828efb545e2742fc05dbd5e17fb168123d5953/parser.py#L319-L325

ngjunsiang:
# Implementing pass-by-value

Our current resolver code certainly isn't using pass-by-value; it's still `verify()`ing the procedure's `stmts` with the global frame 😅:
https://github.com/nyjc-computing/pseudo/blob/7e828efb545e2742fc05dbd5e17fb168123d5953/resolver.py#L86-L88

Should we create a `local` frame here? How would we pass that to the interpreter?

Each time the procedure is called, we will need:
- `BYREF` (global) args to be copied to `local` (unless they are already there)
- `BYVALUE` args to be declared in `local`
- params to be assigned from args

The resolver feels like a more appropriate place to set up the `local` frame then. But first we'll need to do some tweaking to the procedure parser, to make declaration of `BYVALUE` args easier.

## Handling parameter declarations

For starters, notice how similar parameter declarations are to variable declarations. We have a name, followed by `:`, followed by a type keyword. The key difference is that variable declarations are called through a DECLARE statement, with a line break expected at the end, while parameter declarations are an expression within a PROCEDURE statement.

Let's set up a `declare()` expression parser for parameter declarations, and use it inside `declareStmt()` to minimise code duplication:
https://github.com/nyjc-computing/pseudo/blob/f52c24235e2d71bf047ef3a718c3cc7862f96c65/parser.py#L136-L146

https://github.com/nyjc-computing/pseudo/blob/f52c24235e2d71bf047ef3a718c3cc7862f96c65/parser.py#L148-L156

Since declarations need to be type-checked, we might as well do it in `declare()` and not have to do it anymore in `declareStmt()`.

ngjunsiang:
Let's refactor `procedureStmt()` to use our handy dandy new `declare()` expression parser:

https://github.com/nyjc-computing/pseudo/blob/090821db3047d0dcc471c57bf5de6014edd4d74e/parser.py#L304-L316

Here we fix an early ~mistake~~ decision: instead of using dicts for `params`, since we how have `var` dicts for holding the name and type, let's just use a list instead. Those are so much easier to iterate through.

ngjunsiang:
## Resolving BYREF parameters

First we declare our params as local vars. This is now made easier with our `var` expression (which has a `'name'` and `'type'` key), since we can draw on our earlier work in `verifyDeclare()`:
https://github.com/nyjc-computing/pseudo/blob/63b3d184d9b3a1e213bd26d877132a741133a059/resolver.py#L86-L91

If the procedure is declared with BYREF, reassignments of the parameters cause the global variable to change as well. So we need to type-check BYREF parameters:
https://github.com/nyjc-computing/pseudo/blob/63b3d184d9b3a1e213bd26d877132a741133a059/resolver.py#L92-L101

Not forgetting to verify the procedure's own statements. But we pass the `local` frame instead of the global frame:
https://github.com/nyjc-computing/pseudo/blob/63b3d184d9b3a1e213bd26d877132a741133a059/resolver.py#L102-L104

ngjunsiang:
Our CALL resolver needs updating as well, to handle `params` as a list instead of a dict:
https://github.com/nyjc-computing/pseudo/blob/c9e3d8d37e9f2cd768f665150ebedcc1b6940127/resolver.py#L128-L135

ngjunsiang:
And now fixing our CALL executor:
https://github.com/nyjc-computing/pseudo/blob/72badaaa5b315febaa8dd22277513a5c3d01a4dd/interpreter.py#L81-L87

ngjunsiang:
# Rewind

Oops ... here we realise we won't be able to update BYREF variables; if they are changed in local, how would we know, and how would we change them back in global? Time to rethink our approach from https://github.com/nyjc-computing/pseudo/pull/24#issuecomment-1105930470.

Recall that

> if there are several parameters, they should all be passed by the same method and the `BYVALUE` or `BYREF` keyword need not be repeated.

That suggests a different approach to local variable declaration. Recall that when we declare variables, we add a dict to the frame:
https://github.com/nyjc-computing/pseudo/blob/72badaaa5b315febaa8dd22277513a5c3d01a4dd/resolver.py#L48-L51

For lack of a better name, let's call this dict a **slot**. So the frame is a dict with the name as key, and a slot as value.

When we initialise variables in the procedure's `local` frame, we have a choice:
- Create a new slot in `local` if the variables are pass-by-value
- Use the slot from global frame if the variables are pass-by-reference

Let's see what that looks like.

ngjunsiang:
# Resolving BYREF parameters, redux

We still have a `local` frame:
https://github.com/nyjc-computing/pseudo/blob/d8167c1b0c0817e638052beaec44a1fd5f3a4869/resolver.py#L86-L88

Whether for BYVALUE or BYREF variables, we have to loop through `params`:
https://github.com/nyjc-computing/pseudo/blob/d8167c1b0c0817e638052beaec44a1fd5f3a4869/resolver.py#L89-L93

For BYVALUE variables, we just declare local variables with new slots like we did in the previous version.
For BYREF variables, instead of declaring new slots, we reuse the slot from the global frame:
https://github.com/nyjc-computing/pseudo/blob/d8167c1b0c0817e638052beaec44a1fd5f3a4869/resolver.py#L94-L95

We do the usual type-checking (param type should match global var type) ...
https://github.com/nyjc-computing/pseudo/blob/d8167c1b0c0817e638052beaec44a1fd5f3a4869/resolver.py#L96-L101

Before we *reference* the global frame slot in the `local` frame:
https://github.com/nyjc-computing/pseudo/blob/d8167c1b0c0817e638052beaec44a1fd5f3a4869/resolver.py#L102-L103

And the rest of the code is the same. With one small embarrassing fix: we were checking the passby token instead of the word:
[16ce3cdae838f59b992b743d27ded30da8723c11]

ngjunsiang:
# Testing

Code:
```
DECLARE Person : STRING
PROCEDURE SayHi(BYREF Person : STRING)
    OUTPUT "Hi, ", Person, "!"
ENDPROCEDURE
CALL SayHi("John")
```

Result:
```
Expected variable name, got ':'
```

That's because of these two lines:
https://github.com/nyjc-computing/pseudo/blob/16ce3cdae838f59b992b743d27ded30da8723c11/parser.py#L309-L310

`match()` already consumes a token, and `consume()` consumes another one. Fix: [9d051a97d9d261798bf45c788a4c1beae3c62540]

Result:
```
Expect STRING for BYREF Person, got {'type': 'name', 'word': 'STRING', 'value': None}
```

That's because each `var` in `stmt['params']` still contains the type token, which needs to be `resolve()`d to get the type string. (Confused? Me too. This is screaming for a refactor. But not yet.) The fix: [20ff6599d690512d60b6d6a94850ce6b86bf4430]

Result:
```
{ 'Person': {'type': 'STRING', 'value': 'John'},
  'SayHi': { 'type': 'procedure',
             'value': { 'frame': { 'Person': { 'type': 'STRING',
                                               'value': 'John'}},
                        'params': [ { 'name': { 'type': 'name',
                                                'value': None,
                                                'word': 'Person'},
                                      'type': { 'type': 'name',
                                                'value': None,
                                                'word': 'STRING'}}],
                        'passby': 'BYREF',
                        'stmts': [ { 'exprs': [ { 'type': 'string',
                                                  'value': 'Hi, ',
                                                  'word': '"Hi, "'},
                                                { 'left': { 'Person': { 'type': 'STRING',
                                                                        'value': 'John'}},
                                                  'oper': { 'type': 'symbol',
                                                            'value': <function get at 0x7f257f47d0d0>,
                                                            'word': ''},
                                                  'right': { 'type': 'name',
                                                             'value': None,
                                                             'word': 'Person'}},
                                                { 'type': 'string',
                                                  'value': '!',
                                                  'word': '"!"'}],
                                     'rule': 'output'}]}}}
```

ngjunsiang:
# Interpreting BYREF calls

With our `local` frame now set up correctly, we can use it directly from the procedure stored in the global frame:
https://github.com/nyjc-computing/pseudo/blob/213ff6c05c64bcd36e55abee2576c84ace18ff91/interpreter.py#L79-L85

This way, BYREF variables will be accessed directly from the global variable slot, while BYVALUE variables will be ... stored in the same local frame?! That sounds dangerous ... shouldn't each call use a fresh frame for BYVALUE variables?

Well yes ... we don't want to complicate our code more than it needs to be though. The next step is to assign arguments into the `local` frame. Since 9608 pseudocode does not support default argument values, all the local frame variables will be overridden anyway. We have to be careful not to evaluate any `get`s from `local` frame before doing so.

And the rest of the code remains the same:
https://github.com/nyjc-computing/pseudo/blob/213ff6c05c64bcd36e55abee2576c84ace18ff91/interpreter.py#L87-L93

Elegantly, `local[name]` passes the global variable slot, allowing values to be directly `get`ted from or assigned to the global frame.

ngjunsiang:
# Testing

Code:
```
DECLARE Person : STRING
PROCEDURE SayHi(BYREF Person : STRING)
    OUTPUT "Hi, ", Person, "!"
    Person <- "Mary"
ENDPROCEDURE
CALL SayHi("John")
OUTPUT Person
```

Result:
```
Hi, John!
Mary
```

Our BYREF works 😎

PR #25, opened 2022-04-22
13a Functions

ngjunsiang:
The previous chapter, [Procedures](https://github.com/nyjc-computing/pseudo/pull/20), introduced us to many new concepts that we will need here: use of a local frame, as well as passing by value.

The 9608 pseudocode reference guide says the following for functions:

> Functions operate in a similar way to procedures, except that in addition they return a single value to the
> point at which they are called. Their definition includes the data type of the value returned.
> 
> A procedure with no parameters is defined as follows:
> ```
> FUNCTION <identifier> RETURNS <data type>
>  <statements>
> ENDFUNCTION
> ```
> A procedure with parameters is defined as follows:
> ```
> FUNCTION <identifier>(<param1>:<datatype>,<param2>:<datatype>...) RETURNS <data type>
> <statements>
> ENDFUNCTION
> ```
> 
> The keyword RETURN is used as one of the statements within the body of the function to specify the value to
be returned. Normally, this will be the last statement in the function definition.
>
> Because a function returns a value that is used when the function is called, function calls are not complete program statements. The keyword CALL should not be used when calling a function. Functions should only be called as part of an expression. When the RETURN statement is executed, the value returned replaces the function call in the expression and the expression is then evaluated.
> 
> **Example – definition and use of a function**
> ```
> FUNCTION Max(Number1:INTEGER, Number2:INTEGER) RETURNS INTEGER
>     IF Number1 > Number2
>       THEN
>         RETURN Number1
>       ELSE
>         RETURN Number2
>     ENDIF
> ENDFUNCTION
> OUTPUT "Penalty Fine = ", Max(10,Distance*2)
> ```
> 
> In principle, parameters can also be passed by value or by reference to functions and will operate in a similar way. However, it should be considered bad practice to pass parameters by reference to a function and this should be avoided. Functions should have no other side effect on the program other than to return the designated value.

What this means for us:
- We'll have to figure out how to parse a FUNCTION statement
- We'll have to implement a RETURN statement
- We'll have to type-check function returns with their declared type
- We'll have to figure out how to `evaluate()` a function as an expression with a resulting value, instead of execute()`ing it as a statement
- We'll assume function parameters can only be defined BYVALUE, and treat any attempts to define them BYREF as a `ParseError`

That's four mouthfuls to chew. One at a time.

### DIFF
--- a/builtin.py
+++ b/builtin.py
@@ -43,6 +43,9 @@ def eq(x, y):
 def get(frame, name):
     return frame[name]['value']
 
+def call(func, args):
+    return func, args
+
 
 
 # Token types
--- a/interpreter.py
+++ b/interpreter.py
@@ -1,3 +1,4 @@
+from builtin import call
 from builtin import RuntimeError
 
 
@@ -13,6 +14,12 @@ def evaluate(frame, expr):
         return expr
     # Evaluating exprs
     oper = expr['oper']['value']
+    if oper is call:
+        left = evaluate(frame, expr['left'])
+        right = expr['right']
+        func, args = oper(left, right)
+        # ...
+        # return result
     left = evaluate(frame, expr['left'])
     right = evaluate(frame, expr['right'])
     return oper(left, right)
--- a/interpreter.py
+++ b/interpreter.py
@@ -18,8 +18,19 @@ def evaluate(frame, expr):
         left = evaluate(frame, expr['left'])
         right = expr['right']
         func, args = oper(left, right)
-        # ...
-        # return result
+            # 'frame': local,
+            # 'passby': 'BYVALUE',
+            # 'params': stmt['params'],
+            # 'stmts': stmt['stmts'],
+        local = func['frame']
+        # Assign args into local with param names
+        for arg, param in zip(args, func['params']):
+            name = evaluate(local, param['name'])
+            local[name]['value'] = evaluate(frame, arg)
+        for funcstmt in func['stmts']:
+            returnval = execute(local, funcstmt)
+            if returnval:
+                return returnval
     left = evaluate(frame, expr['left'])
     right = evaluate(frame, expr['right'])
     return oper(left, right)
--- a/interpreter.py
+++ b/interpreter.py
@@ -144,7 +144,7 @@ def execute(frame, stmt):
     if stmt['rule'] == 'call':
         execCall(frame, stmt)
     if stmt['rule'] == 'return':
-        execReturn(frame, stmt)
+        return execReturn(frame, stmt)
 
 def interpret(statements, frame=None):
     if frame is None:
--- a/interpreter.py
+++ b/interpreter.py
@@ -66,6 +66,9 @@ def execRepeat(frame, stmt):
 def execProcedure(frame, stmt):
     pass
 
+def execFunction(frame, stmt):
+    pass
+
 def execCall(frame, stmt):
     # frame[name] = {
     #     'type': 'procedure',
@@ -111,6 +114,8 @@ def execute(frame, stmt):
         execRepeat(frame, stmt)
     if stmt['rule'] == 'procedure':
         execProcedure(frame, stmt)
+    if stmt['rule'] == 'function':
+        execFunction(frame, stmt)
     if stmt['rule'] == 'call':
         execCall(frame, stmt)
 
--- a/interpreter.py
+++ b/interpreter.py
@@ -18,10 +18,6 @@ def evaluate(frame, expr):
         left = evaluate(frame, expr['left'])
         right = expr['right']
         func, args = oper(left, right)
-            # 'frame': local,
-            # 'passby': 'BYVALUE',
-            # 'params': stmt['params'],
-            # 'stmts': stmt['stmts'],
         local = func['frame']
         # Assign args into local with param names
         for arg, param in zip(args, func['params']):
@@ -88,15 +84,6 @@ def execFunction(frame, stmt):
     pass
 
 def execCall(frame, stmt):
-    # frame[name] = {
-    #     'type': 'procedure',
-    #     'value': {
-    #         'frame': local,
-    #         'passby': str,
-    #         'params': stmt['params'],
-    #         'stmts': stmt['stmts'],
-    #     }
-    # }
     # Get procedure from frame
     proc = evaluate(frame, stmt['name'])
     # Note: for BYREF variables, the procedure's frame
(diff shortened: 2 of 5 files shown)


ngjunsiang:
# Parsing function statements

Much of the work has been done in parsing procedures. We'll use most of that code, take out the BYREF/BYVALUE handling.

And we add in some code to handle the RETURNS:
https://github.com/nyjc-computing/pseudo/blob/283723f8ba535d4e2f46d9d5ce74d51b66748e0e/parser.py#L361-L365

And add that to our function `stmt`:
https://github.com/nyjc-computing/pseudo/blob/de19a53a2bd6eeb73275c87bb3dcc9da3d92a0e1/parser.py#L371-L378

ngjunsiang:
# Parsing RETURN statements

RETURN statements aren't too hard to parse:
https://github.com/nyjc-computing/pseudo/blob/8b6e326bfb4a067d70f637bc61ae185668d5ace1/parser.py#L381-L388

Let's test that:
```
DECLARE Five : INTEGER
FUNCTION AddOne(Num : INTEGER) RETURNS INTEGER
    RETURN Num + 1
ENDFUNCTION
Five <- 5
OUTPUT "5 + 1 is ", AddOne(Five)
```

Result:
```
Expected \n
```

We haven't implemented parsing function calls yet ...

ngjunsiang:
# Parsing function calls

Functions can be called anywhere a value or expression is expected. This is parsed by `value()`; that's the natural place to start adding code to handle function calls.

The first token in a function call is the function name: this is a `'name'`-type token, which gets parsed as a `get` expression:
https://github.com/nyjc-computing/pseudo/blob/17826c09512831a37e47811cae4c927c8464f515/parser.py#L53-L56

Let's check for a `'('` after that; we also parse any args, and then expect a `')'` (most of this code is repeated from `callStmt()`):
https://github.com/nyjc-computing/pseudo/blob/17826c09512831a37e47811cae4c927c8464f515/parser.py#L57-L64

Hmm ... what kind of expression should we return?

ngjunsiang:
# Function expressions

We made an early decision to represent expressions as binary expressions, with an `'oper'`, `'left'`, and `'right'`. Should we stick with this format or make a new type of expression?

A `get` operator won't be enough for us to `evaluate()` a function value. And our interpreter is going to `return oper(left, right) in `evaluate()` ...should we make a `call()` operator? But this operator would need to know about `expr`s and `stmt`s, so putting it in `builtin.py doesn't feel appropriate.

It looks like we do need a new kind of expression after all; a `call` expression. It's kind of difficult to think about how to `evaluate()` a `call` expression before we've even `resolve()`d a FUNCTION statement. Let's do that first.

ngjunsiang:
# Resolving FUNCTION statements

This is going to look very similar to resolving PROCEDURE statements, but with `'passby'` always being `'BYVALUE'`, and with an extra `'returns'` key:
https://github.com/nyjc-computing/pseudo/blob/857be1acad4643fd4da7022d8c379c136058f3e2/resolver.py#L145-L165

ngjunsiang:
# Call expressions

A call involves a function, and its arguments; that's two things. Still feels very binary. It would be a waste to make a new kind of expression, with all the complexity that adds.

Let's make a simple `call` operator that doesn't need to know about tokens, functions, or statements:
https://github.com/nyjc-computing/pseudo/blob/f70800dd08cfcee745fff04c172500e610168cba/builtin.py#L46-L47

And construct a function call as a binary expression with a `call` operator:
https://github.com/nyjc-computing/pseudo/blob/4d4f0d88bdea02ef26f9f5068b318c777250f51f/parser.py#L60-L68

`'left'` is a get expression for the function, and `'right'` is the list of arguments. (`resolve()` is not going to like that list of args ...)

ngjunsiang:
# Resolving call expressions

Our function expression is going to be handled by `resolve()` in the resolver, so let's make it work.
https://github.com/nyjc-computing/pseudo/blob/947b5f0fdc053fa41360c751a66541b2d15aa8a4/resolver.py#L26-L32

Our work in [aaaac2a242603b2267add29ab8c4c0411b57dbe0] made this really easy.

ngjunsiang:
# Executing function declarations

Hah! That's done in the resolver already:
https://github.com/nyjc-computing/pseudo/blob/55bdc42bf3c6545712f211fb331c10e149a83e88/interpreter.py#L69-L70

ngjunsiang:
# Evaluating function calls

Okay this is the tricky part. Up to this point, `evaluate()` has always been able to just `return oper(left, right)`. But that's not going to be the case for `call()`; the poor statement trying to `evaluate()` a `call` expression would end up with a `func, args` tuple.

So let's add some conditional evaluation:
https://github.com/nyjc-computing/pseudo/blob/f46fbcc1857139fd3fa85daf44e57c627f2dac64/interpreter.py#L17-L22

Hold on ... we forgot about RETURN statements! (We need to type-check RETURN statements against the function RETURNS.)

ngjunsiang:
# Resolving return type

It's easy enough to get the return type with the help of `resolve()`:
https://github.com/nyjc-computing/pseudo/blob/99a3466d5fc801bc9606f42febe0e278261ffad0/resolver.py#L172-L176

We don't expect to be getting RETURN statements anywhere but in a function, so we can assume `local` as the frame. We `resolve()` the return expression to get its type.

Now we need to capture that return type and use it somewhere:
https://github.com/nyjc-computing/pseudo/blob/f44c8ca5cca33c519ba08c7e90ac39661e5d134e/resolver.py#L204-L205

Notice that this is the only verifier that returns something. Everything else returns `None`. Let's type-check that return value within `verifyFunction()`:
https://github.com/nyjc-computing/pseudo/blob/60898e91010d9176ba3aba517532c4c7a71b5c4e/resolver.py#L160-L163

ngjunsiang:
# Back to evaluating function calls

Let's execute RETURN statements:
https://github.com/nyjc-computing/pseudo/blob/1da2198722083b1979639dae922ce2315296ed31/interpreter.py#L116-L120

Now we can capture return values:
https://github.com/nyjc-computing/pseudo/blob/504b5c7afc4f6f1de1d25ee0edf428c914c1c2fe/interpreter.py#L25-L33

ngjunsiang:
# Testing and bugfix

We forgot to `resolve()` function call args: [56dc061da4ab2a3e89636712d26c06d3ea86c328]

And we also forgot to return the return value in `execute()`: [8a8d166802f9247815d2a3cdad267cf57b3a6a21]

Test code:
```
DECLARE Five : INTEGER
FUNCTION AddOne(Num : INTEGER) RETURNS INTEGER
    RETURN Num + 1
ENDFUNCTION
Five <- 5
OUTPUT "5 + 1 is ", AddOne(Five)
```

Result:
```
5 + 1 is 6
```

Annnnnd ... we have working functions 🙌

PR #26, opened 2022-04-22
13b Loose ends

ngjunsiang:
Let's wrap up a few loose ends for functions and procedures before we move on.

### DIFF
--- a/interpreter.py
+++ b/interpreter.py
@@ -3,6 +3,13 @@ from builtin import RuntimeError
 
 
 
+# Helper functions
+
+def assignArgsParams(frame, args, local, params):
+    for arg, param in zip(args, params):
+        name = evaluate(local, param['name'])
+        local[name]['value'] = evaluate(frame, arg)
+
 def evaluate(frame, expr):
     # Evaluating tokens
     if 'type' in expr:
@@ -17,10 +24,7 @@ def evaluate(frame, expr):
     if oper is call:
         func = evaluate(frame, expr['left'])
         local = func['frame']
-        args, params = expr['right'], func['params']
-        for arg, param in zip(args, params):
-            name = evaluate(local, param['name'])
-            local[name]['value'] = evaluate(frame, arg)
+        assignArgsParams(frame, expr['right'], func['frame'], func['params'])
         for funcstmt in func['stmts']:
             returnval = execute(local, funcstmt)
             if returnval:
@@ -88,10 +92,7 @@ def execCall(frame, stmt):
     # Do not evaluate any get exprs for local frame
     # before assigning local variables from args.
     local = proc['frame']
-    args, params = stmt['args'], proc['params']
-    for arg, param in zip(args, params):
-        name = evaluate(local, param['name'])
-        local[name]['value'] = evaluate(frame, arg)
+    assignArgsParams(frame, stmt['args'], proc['frame'], proc['params'])
     for callstmt in proc['stmts']:
         execute(local, callstmt)
 
--- a/interpreter.py
+++ b/interpreter.py
@@ -5,10 +5,10 @@ from builtin import RuntimeError
 
 # Helper functions
 
-def assignArgsParams(frame, args, local, params):
-    for arg, param in zip(args, params):
-        name = evaluate(local, param['name'])
-        local[name]['value'] = evaluate(frame, arg)
+def assignArgsParams(frame, args, callable):
+    for arg, param in zip(args, callable['params']):
+        name = evaluate(callable['frame'], param['name'])
+        callable['frame'][name]['value'] = evaluate(frame, arg)
 
 def evaluate(frame, expr):
     # Evaluating tokens
@@ -24,7 +24,7 @@ def evaluate(frame, expr):
     if oper is call:
         func = evaluate(frame, expr['left'])
         local = func['frame']
-        assignArgsParams(frame, expr['right'], func['frame'], func['params'])
+        assignArgsParams(frame, expr['right'], func)
         for funcstmt in func['stmts']:
             returnval = execute(local, funcstmt)
             if returnval:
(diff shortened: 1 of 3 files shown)


ngjunsiang:
First, a quick bugfix for missing out arguments to `expression()`: [737ef5722744c52e33f6c00c8779db2655768d63]

# Type-checking BYREF arguments

BYREF arguments are going to be assigned into the `local` frame from global frame, which means they must be names and not arithmetic expressions:
[in `verifyCall()`]
https://github.com/nyjc-computing/pseudo/blob/83f00b5b3fa7bcd37d541740951295c041cc76d5/resolver.py#L149-L153

ngjunsiang:
Actually, this type-checking makes sense for functions too, if ever we decide to allow BYREF in them ...

It feels like the bigger issue here is that function calls and procedure calls are essentially the same, even though the former is an expression evaluation and the latter is a statement execution. They should be able to share some code, dammit!

We pulled off something like this in `parser.py`, where we made a `declare()` function for parsing declaration expressions (NAME : TYPE) and used them for declaration statements, procedure statements, and function statements:
https://github.com/nyjc-computing/pseudo/blob/83f00b5b3fa7bcd37d541740951295c041cc76d5/parser.py#L148-L158

Let's see if we can do that for `resolver.py` and `interpreter.py`.

ngjunsiang:
# Refactoring function and parameter resolving

Both involve:
1. Setting up a `local` frame
2. Declaring vars in `local` frame
3. BYREF arguments: type-checking `local` against global frame vars
4. Resolve statements
5. Declare function/procedure in global frame

This declaration thing seems to happen a lot; `verifyProcedure()` and `verifyFunction()` even use `verifyDeclare()`:
https://github.com/nyjc-computing/pseudo/blob/83f00b5b3fa7bcd37d541740951295c041cc76d5/resolver.py#L100-L103
https://github.com/nyjc-computing/pseudo/blob/83f00b5b3fa7bcd37d541740951295c041cc76d5/resolver.py#L160-L165

Let's make type-checking easier with a helper function:
https://github.com/nyjc-computing/pseudo/blob/d20541c63f51aa9a7c91ae632daff8bc84b6220d/resolver.py#L10-L13

ngjunsiang:
Let's make type-checking easier with a helper function:
https://github.com/nyjc-computing/pseudo/blob/d20541c63f51aa9a7c91ae632daff8bc84b6220d/resolver.py#L10-L13

And that helps us clean up 20 lines of code: [f1b0b028c68d7c55e7bcf1d478ea2ef9b861b16f]

ngjunsiang:
# Loop helpers

Another common pattern is looping through a list and `verify()`ing or `resolve()`ing:
https://github.com/nyjc-computing/pseudo/blob/f1b0b028c68d7c55e7bcf1d478ea2ef9b861b16f/resolver.py#L82-L83
https://github.com/nyjc-computing/pseudo/blob/f1b0b028c68d7c55e7bcf1d478ea2ef9b861b16f/resolver.py#L53-L54

It's not a major annoyance, but detracts from the code-reading; it isn't at all difficult to make simple helpers for them:
https://github.com/nyjc-computing/pseudo/blob/8dbdae192365055f416b5c643d2c4702a9393b3d/resolver.py#L15-L21

ngjunsiang:
# Function and procedure calls

In `interpreter.py`: we have `execCall()` for procedure calls, and `evaluate()` for function calls. Both share quite similar code:
https://github.com/nyjc-computing/pseudo/blob/cbf206b555e99a69aa80a02dbc45ff2e7beca932/interpreter.py#L17-L29
https://github.com/nyjc-computing/pseudo/blob/cbf206b555e99a69aa80a02dbc45ff2e7beca932/interpreter.py#L86-L101

Let's do a rewrite to make them easier to refactor: [ec9916e89bd289a6fccbfc816c9656c32581e967]

ngjunsiang:
# Assigning arguments to parameters

Let's make a helper to clarify the bulk of the complexity: assigning arguments to parameters
https://github.com/nyjc-computing/pseudo/blob/41d82e1864efb177e3ce6b1b525f34e9e59f62c3/interpreter.py#L8-L11

And 4 lines of code reduces to 2: [41d82e1864efb177e3ce6b1b525f34e9e59f62c3]

Can we go further? Yes we can!
https://github.com/nyjc-computing/pseudo/blob/41d82e1864efb177e3ce6b1b525f34e9e59f62c3/interpreter.py#L27
https://github.com/nyjc-computing/pseudo/blob/41d82e1864efb177e3ce6b1b525f34e9e59f62c3/interpreter.py#L95

The last two arguments seem unnecessary ... the `local` frame and `params` are both coming from the stored function/procedure. For lack of a better name, and because this name is so widely used, let's call them a callable:
https://github.com/nyjc-computing/pseudo/blob/23639d33b1f196e5e4c71b12936581f7120e2371/interpreter.py#L8-L11

And now we can assign arguments from global frame to parameters in a `local` frame with a helper function call, and three arguments:
23639d33b1f196e5e4c71b12936581f7120e2371

ngjunsiang:
# Helping loop execution

Like we did with `verifyStmts()` earlier in `resolver.py`, let's make an `executeStmts()` for our interpreter:
https://github.com/nyjc-computing/pseudo/blob/6e4affd20eefaf3d75c9fd4d95c6f07333302623/interpreter.py#L8-L10

It's got to handle return statements from functions though. Let's upgrade it:
https://github.com/nyjc-computing/pseudo/blob/420d67f5d25b2d8f2f9caf1e33dc7f26fe2dd38c/interpreter.py#L8-L12

And then use it. Turns out that `interpreter.py` is full of loop executions!
[31b4fde4b7c53ebdd9c38a682a6737854926ea4e]

ngjunsiang:
# Caveat

Something we've got to watch out for: `executeStmts()` will return once a return value is received, and skip further statements. This should only be done in a function, but we didn't check for that. Thus far ... we haven't put in any code to stop users (i.e. pseudocode programmers) from using RETURN statements elsewhere. We will have to implement these safeguards later.

Let's add an issue for it: #27

ngjunsiang:
And we're done with the bulk of it, actually!

There's File IO, which we can draw on Python for, and built-in functions, which we now have the infrastructure for, having implemented variables and functions.

PR #28, opened 2022-04-23
14a Reading from source

ngjunsiang:
Before we embark on File IO, it feels kind of strange to do so when our developers have to edit *the source code* of our main program to insert their code ... frankly it would be embarrassing to see this in real-world code!

So let's at least put something in place for us to read our source code from an external source file. For now, we'll just use `main.pseudo` as the filename of this source file (we don't support importing code from multiple source files anyway).

And let's read in the source code from that file at the start:
https://github.com/nyjc-computing/pseudo/blob/8d87a0983ec887b1e16d5fb4bac5fd6668bfd206/main.py#L12-L14

### DIFF
--- a/main.pseudo
+++ b/main.pseudo
@@ -1,0 +1,6 @@
+DECLARE Five : INTEGER
+FUNCTION AddOne(Num : INTEGER) RETURNS INTEGER
+    RETURN Num + 1
+ENDFUNCTION
+Five <- 5
+OUTPUT "5 + 1 is ", AddOne(Five)
--- a/main.py
+++ b/main.py
@@ -9,18 +9,9 @@ import interpreter
 
 
 
-src = '''
-DECLARE Five : INTEGER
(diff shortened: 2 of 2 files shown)


ngjunsiang:
What if the user has a different file they want to use? It's common to pass that as the first argument to our interpreter as follows:

`main.py myotherfile.pseudo`

(Right now our file is called `main.py`. We'll give it a cooler name when we're packaging it.)

In python, we can extract these arguments passed to our program using `sys.argv`, which gives a list of arguments. The first argument is always the name of the file (in our case, `main.py`). So any filename passed to our program would be the second argument:

https://github.com/nyjc-computing/pseudo/blob/0be6bb23111255a1e9c9fd9087e70169bab771b9/main.py#L15-L18

ngjunsiang:
Now that's looking more respectable.

On to the next thing a proper compiler needs: error messages.

PR #29, opened 2022-04-23
14b Line numbers

ngjunsiang:
The first thing that error messages have, which we don't, are line numbers. For example:

```
  File "resolver.py", line 17
    resolve(frame, expr)
    ^
IndentationError: expected an indented block
```

We don't have the infrastructure to show lines of code yet, but line numbers we can certainly do!

### DIFF
--- a/builtin.py
+++ b/builtin.py
@@ -10,6 +10,9 @@ class PseudoError(Exception):
         else:
             self.line = token['line']
 
+    def msg(self):
+        return self.args[0]
+
 class ParseError(PseudoError):
     """Custom error raised by scanner and parser."""
     def report(self):
@@ -17,7 +20,7 @@ class ParseError(PseudoError):
             token = self.token['word']
         else:
             token = self.token
-        return f"[{self.line}] {repr(token)}: {self.args[0]}"
+        return f"[{self.line}] {repr(token)}: {self.msg()}"
 
 class RuntimeError(Exception):
     """Custom error raised by interpreter."""
--- a/builtin.py
+++ b/builtin.py
@@ -20,7 +20,7 @@ class ParseError(PseudoError):
             token = self.token['word']
         else:
             token = self.token
-        return f"[{self.line}] {repr(token)}: {self.msg()}"
+        return f"[Line {self.line}] {repr(token)}: {self.msg()}"
 
 class RuntimeError(Exception):
     """Custom error raised by interpreter."""
--- a/builtin.py
+++ b/builtin.py
@@ -2,10 +2,13 @@
 
 class PseudoError(Exception):
     """Base exception class for all Psuedo errors."""
-    def __init__(self, msg, token):
+    def __init__(self, msg, token, line=None):
         super().__init__(msg)
         self.token = token
-        self.line = token['line']
+        if line is not None:
+            self.line = line
+        else:
+            self.line = token['line']
 
 class ParseError(PseudoError):
     """Custom error raised by scanner and parser."""
--- a/builtin.py
+++ b/builtin.py
@@ -1,10 +1,22 @@
 # Errors
 
-class ParseError(Exception): pass
+class ParseError(Exception):
+    def __init__(self, msg, token):
+        super().__init__(msg)
+        self.token = token
+        self.line = token['line']
 
-class RuntimeError(Exception): pass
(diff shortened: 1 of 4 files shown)


ngjunsiang:
# Tracking line numbers

Let's start by tracking line numbers in `code` in our scanner (good thing we decided to wrap it in a dict!)
https://github.com/nyjc-computing/pseudo/blob/4146603a031a0d393a80ccb875e7202e80300b08/scanner.py#L60-L65

And increment that counter whenever we reach a line break:
https://github.com/nyjc-computing/pseudo/blob/4146603a031a0d393a80ccb875e7202e80300b08/scanner.py#L71-L74

ngjunsiang:
# Adding line numbers

Our `makeToken()` helpers gain a `'line'` key:
https://github.com/nyjc-computing/pseudo/blob/421ffddc66351648ff8927b65ad91629672d68ba/scanner.py#L19-L25

Everywhere we make a token, we make a best-guess effort at adding line numbers to the token. This isn't a problem in our scanner, which has access to the `code` dict.

In the parser, however, we insert our own tokens, like the `EOF` token at the end, and when we desugar syntax or add operators. For now we use the line number of the nearest token, and come back to refine it later.

These changes are shown in [421ffddc66351648ff8927b65ad91629672d68ba].

ngjunsiang:
# Reporting line numbers

Our error messages are caught in `main.py`, where they are printed. So we need a way to pass these line numbers up. My first shot would be to add them in the raised error.

Since the line number comes from the token, we might as well add the whole token to the error:
https://github.com/nyjc-computing/pseudo/blob/8303f36fa9a827eeb1e12d4538a093469752f9a6/builtin.py#L3-L19

This is senseless duplication of code. While the focus of these chapters is not on OOP, applying inheritance here makes the code much simpler. I'll also take the chance to add some long-needed docstrings:
https://github.com/nyjc-computing/pseudo/blob/84a95c57d0a7aea8d27f4a5b70fcc0231f1a09f1/builtin.py#L3-L17

ngjunsiang:
Oops! We forgot that the scanner scans characters and not tokens; `ParseError`s raised there would not have a token.

No issue, we'll just make the `token` argument serve double-duty as either a `str` or a `token` (for now). And we have `line` as an optional third argument:
https://github.com/nyjc-computing/pseudo/blob/4c30e537a08f0d663782bcdb0b7b69653f40d436/builtin.py#L3-L11

ngjunsiang:
Then let's start adding tokens and line numbers to our raised errors. Since the error contains token information, we can take it out of the f-strings and re-word some of the errors.
https://github.com/nyjc-computing/pseudo/blob/59023c57d839963030b348190636ca2a143be447/scanner.py#L93-L97

Let's make a `report()` method to get a nice formatted string. Along the way, I also make a `msg()` method for the base error class to get the error message more easily:
(In class `PsuedoError`)
https://github.com/nyjc-computing/pseudo/blob/9e33ba491c92c019cd12cea074527f784f937b38/builtin.py#L13-L14

https://github.com/nyjc-computing/pseudo/blob/9e33ba491c92c019cd12cea074527f784f937b38/builtin.py#L16-L23

ngjunsiang:
And we make the changes in `main.py` to use the freshly written `report()` method:
https://github.com/nyjc-computing/pseudo/blob/efce9e1274c1c869adad7b1c0e98670a23dbe3bb/main.py#L24-L25

Result:
`[1] ';': Unrecognised character`

ngjunsiang:
On to adding tokens and lines to errors in the parser:
https://github.com/nyjc-computing/pseudo/blob/e127f02bd79b37cbae46862208a06c47ffc991e4/parser.py#L29-L33

and the rest in [c95852171c346f215aa386c18a8a6c0388c99f16]

ngjunsiang:
Let's make better parser error messages while we can.

...

ngjunsiang:
# Bugfixes

Earlier in [c0ded13d0b241231aba0712da6f9539ad3484182] I accidentally introduced some bugs; match()` will `consume()` the operator token when that is not what we wanted. This is fixed in [02c66037b3d0f327553617f9b7abdf3b704b23e6].

ngjunsiang:
And with this, we have line numbers and better error messages.

PR #30, opened 2022-04-26
14c Referencing source code

ngjunsiang:
Now that we have line numbers, let's see if we can get the source code to display in our error messages 😎 

That means a pretty major change in the way we parse the code: Instead of `consume()`ing the code while stripping it character by character, we want to keep it intact so we can reference it later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,3 +26,4 @@ It is also an attempt to write a programming book in a new style. Each chapter o
 - [13a Functions](https://github.com/nyjc-computing/pseudo/pull/25)
 - [13b Loose ends](https://github.com/nyjc-computing/pseudo/pull/26)
 - [14a Reading from source](https://github.com/nyjc-computing/pseudo/pull/28)
+- [14b Line numbers](https://github.com/nyjc-computing/pseudo/pull/29)
--- a/builtin.py
+++ b/builtin.py
@@ -13,8 +13,6 @@ class PseudoError(Exception):
     def msg(self):
         return self.args[0]
 
-class ParseError(PseudoError):
-    """Custom error raised by scanner and parser."""
     def report(self):
         if type(self.token) is dict:
             token = self.token['word']
@@ -22,14 +20,14 @@ class ParseError(PseudoError):
             token = self.token
         return f"{repr(token)}: {self.msg()}"
 
+class ParseError(PseudoError):
+    """Custom error raised by scanner and parser."""
+
 class RuntimeError(PseudoError):
     """Custom error raised by interpreter."""
 
 class LogicError(PseudoError):
     """Custom error raised by resolver."""
-    def report(self):
-        token = self.token['word']
-        return f"{repr(token)}: {self.msg()}"
 
 
 
--- a/builtin.py
+++ b/builtin.py
@@ -22,10 +22,10 @@ class ParseError(PseudoError):
             token = self.token
         return f"[Line {self.line}] {repr(token)}: {self.msg()}"
 
-class RuntimeError(Exception):
+class RuntimeError(PseudoError):
     """Custom error raised by interpreter."""
 
-class LogicError(Exception):
+class LogicError(PseudoError):
     """Custom error raised by resolver."""
 
 
--- a/builtin.py
+++ b/builtin.py
@@ -27,6 +27,9 @@ class RuntimeError(PseudoError):
 
 class LogicError(PseudoError):
     """Custom error raised by resolver."""
+    def report(self):
+        token = self.token['word']
+        return f"[Line {self.line}] {repr(token)}: {self.msg()}"
 
 
 
--- a/builtin.py
+++ b/builtin.py
@@ -20,7 +20,7 @@ class ParseError(PseudoError):
             token = self.token['word']
         else:
             token = self.token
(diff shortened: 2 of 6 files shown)


ngjunsiang:
# Scanning with a cursor

> **[cursor](https://www.etymonline.com/word/cursor#etymonline_v_494) (n.)**
>
> c. 1300 (as a surname) "a running messenger," from Latin _cursor_ "runner," also "errand-boy," from _curs-_, past-participle stem of _currere_ "to run" (from PIE root [*kers-](https://www.etymonline.com/word/*kers-?ref=etymonline_crossreference) "to run"). From 1590s as "part of aslide rule or other instrument that slides backward and forward upon another part." The computer screen sense is a 1967 extension of this.

Instead of using the first remaining character in `code['src']`, we use an index to keep track of which character in the source we are currently at.

The technical term for this index is the **cursor**, which "runs along" the code as we scan it, pointing to the next character to check/scan.

Instead of stripping the first character, we advance the cursor:
https://github.com/nyjc-computing/pseudo/blob/1166e3f06bed2744026d2b57645e7e3e387f0edf/scanner.py#L14-L17

This has the added advantage of being more computationally efficient (since we don't have to reassign a string).

We use the cursor to check the current character:
https://github.com/nyjc-computing/pseudo/blob/1166e3f06bed2744026d2b57645e7e3e387f0edf/scanner.py#L11-L12

And we check whether we are at the end of the code by checking the cursor against the code length:
https://github.com/nyjc-computing/pseudo/blob/1166e3f06bed2744026d2b57645e7e3e387f0edf/scanner.py#L8-L9

The following changes are needed in our `code` dict for this to work:
https://github.com/nyjc-computing/pseudo/blob/1166e3f06bed2744026d2b57645e7e3e387f0edf/scanner.py#L62-L67

ngjunsiang:
# Referencing lines

To be able to reference lines, we need to have our scanner pass a list of lines back to `main.py`. Let's add more `code` keys to keep track of lines, and also another cursor to keep track of the start of the current line:
https://github.com/nyjc-computing/pseudo/blob/b7a4c8c3a44c0ac9afdaf0f256dee5259073ba38/scanner.py#L62-L68

Whenever we hit a line break, we string-slice the source code to grab the current line. Then we update the current line cursor:
https://github.com/nyjc-computing/pseudo/blob/b7a4c8c3a44c0ac9afdaf0f256dee5259073ba38/scanner.py#L76-L82

Not forgetting it pass the lines back to `main.py`:
https://github.com/nyjc-computing/pseudo/blob/b7a4c8c3a44c0ac9afdaf0f256dee5259073ba38/scanner.py#L106
https://github.com/nyjc-computing/pseudo/blob/b7a4c8c3a44c0ac9afdaf0f256dee5259073ba38/main.py#L21

ngjunsiang:
# Bugfixes

Minor bugfixes: a typo in an f-string [bab05beb45dd72e57f06394c617ac544dddcccbc]

And fixing incorrect detection of the source code line length: [bde81f9f77973396237a2f250469845062b48c5f]

ngjunsiang:
Let's extend our `report()` method to other types of errors as well, by promoting it to the `PseudoError` parent class instead of keeping it only in `ParseError`:
https://github.com/nyjc-computing/pseudo/blob/8797322dc112be90bfe4122fee2b92916bf95144/builtin.py#L3-L21

We also took out the line info that used to be displayed by the error's `report()` method; it is now displayed in `main.py`:
https://github.com/nyjc-computing/pseudo/blob/cd2bc7170a3bb34a138946e13970c7d7ebfe9436/main.py#L24-L27

And now we make `LogicError` inherit from `PseudoError`:
https://github.com/nyjc-computing/pseudo/blob/6451196f34ad8aae2a20b47897610a2405fb1ca2/builtin.py#L28-L29

And add token information (best guess) to all our uses of `LogicError` in the resolver: [e6ee66d66e61086bfa3a8e919c26dc4519687834]

Test code:
```
DECLARE Five : INTEGER
FUNCTION AddOne(Num : INTEGER) RETURNS INTEGER
    RETURN Num + 1
ENDFUNCTION
Five <- "Five"
OUTPUT "5 + 1 is ", AddOne(Five)
```

Result:
```
[Line 5] Five <- "Five"
'"Five"': Expected 'INTEGER', got 'STRING'
```

ngjunsiang:
# Adding token information to LogicError

Now we make `LogicError` inherit from `PseudoError`:
https://github.com/nyjc-computing/pseudo/blob/6451196f34ad8aae2a20b47897610a2405fb1ca2/builtin.py#L28-L29

And add token information (best guess) to all our uses of `LogicError` in the resolver: [e6ee66d66e61086bfa3a8e919c26dc4519687834]

ngjunsiang:
# Testing

Test code:
```
DECLARE Five : INTEGER
FUNCTION AddOne(Num : INTEGER) RETURNS INTEGER
    RETURN Num + 1
ENDFUNCTION
Five <- "Five"
OUTPUT "5 + 1 is ", AddOne(Five)
```

Result:
```
[Line 5] Five <- "Five"
'"Five"': Expected 'INTEGER', got 'STRING'
```

ngjunsiang:
We could do the same with `RuntimeError` ... but we haven't needed to `raise` those yet. Let's get to that when we need to.

PR #31, opened 2022-04-26
14d Column info

ngjunsiang:
Let's take a short detour and add column info to our errors as well, shall we?

We start by adding column info to our `makeToken()` helper:
https://github.com/nyjc-computing/pseudo/blob/add59be942625e76039b6d930db36787a66b2076/scanner.py#L19-L26

We calculate the column number by subtracting `code['lineStart']` from `code['curr']`:
https://github.com/nyjc-computing/pseudo/blob/320b9c7a145f371e1fb26196ef94834a06589e8f/scanner.py#L81-L87

And we add that info everywhere we call `makeToken()`:
320b9c7a145f371e1fb26196ef94834a06589e8f

### DIFF
--- a/builtin.py
+++ b/builtin.py
@@ -9,6 +9,7 @@ class PseudoError(Exception):
             self.line = line
         else:
             self.line = token['line']
+        self.col = token['col']
 
     def msg(self):
         return self.args[0]
--- a/main.py
+++ b/main.py
@@ -22,7 +22,10 @@ def main():
         statements = parser.parse(tokens)
         statements, frame = resolver.inspect(statements)
     except (ParseError, LogicError) as err:
-        print(f"[Line {err.line}]", lines[err.line - 1])
+        lineinfo = f"[Line {err.line}]"
+        print(lineinfo, lines[err.line - 1])
+        leftmargin = len(lineinfo) + 1 + err.col
+        print((' ' * leftmargin) + '^')
         print(err.report())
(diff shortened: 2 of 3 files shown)


ngjunsiang:
# Adding column info to errors

We initialise our parent `PseudoError` class with the column info:
https://github.com/nyjc-computing/pseudo/blob/add59be942625e76039b6d930db36787a66b2076/builtin.py#L3-L12

ngjunsiang:
# Displaying column info

How should we display this column info in our error messages? We could add column numbers next to line numbers, but then our users (i.e. pseudocode developers) would have to count characters ... not friendly at all!

Let's add a visual indicator instead:
https://github.com/nyjc-computing/pseudo/blob/9759923a882d8e1aaa0fb09ba04aa9707a0371b5/main.py#L24-L30

Result:
```
[Line 5] Five <- "5"
                    ^
'"5"': Expected 'INTEGER', got 'STRING'
```

ngjunsiang:
# Bugfix

Oops ... when we calculated column position, we included the length of the token itself. But usually we want the token to point to the **start** of the token not its end.

Let's fix that: [73e89a904ce264424ad6a49be7d0501658047c38]

Result:
```
[Line 5] Five <- "5"
                 ^
'"5"': Expected 'INTEGER', got 'STRING'
```

ngjunsiang:
And we'll leave it at this :)

PR #32, opened 2022-04-26
15 File IO

ngjunsiang:
We have a pretty complete interpreter at this point!

Before we took a detour to make a presentable interpreter, we were just talking about file IO. Now let's get back to it.


ngjunsiang:
# Recognising IO keywords

Let's update our list of keywords:
https://github.com/nyjc-computing/pseudo/blob/53261dd7b562cb1e93df55e45a061663b841788e/builtin.py#L77-L90

ngjunsiang:
# Test code

Lets use the following psuedocode to test:
```
DECLARE Number : INTEGER
DECLARE Max : INTEGER
Number <- 0
OPENFILE data.txt FOR WRITE
WHILE Number <= Max DO
    WRITEFILE data.txt, Number
CLOSEFILE data.txt
```

ngjunsiang:
# Parsing *file statements

Let's add more rules for our entry point:
https://github.com/nyjc-computing/pseudo/blob/df767f96d931ad54f882e190622de6699d04b73e/parser.py#L438-L445

And define the respective parsers:
https://github.com/nyjc-computing/pseudo/blob/df767f96d931ad54f882e190622de6699d04b73e/parser.py#L392-L411

ngjunsiang:
## Defining *file statements

What kind of `stmt` should we return?

Ideally I hope to avoid needing 4 types of statements, one for each keyword. Let's go with the following:

- rule: file
- action: open | read | write | close
- name: <string>
- mode: read | write | append | None
- data: <string> | None

Now let's implement the parsers.

ngjunsiang:
https://github.com/nyjc-computing/pseudo/blob/4a96955a465277afcb5fbe18d0a57d0e9f91a835/parser.py#L392-L406
Nothing too surprising here. We pass the entire token instead of unpacking it, so that we can raise errors if necessary in the resolver/interpreter.

https://github.com/nyjc-computing/pseudo/blob/4a96955a465277afcb5fbe18d0a57d0e9f91a835/parser.py#L408-L420
READFILE statements take in a *variable name* after the file identifier, and we pass that as the `'data'` key in the `'file'` statement.  READFILE statements do not provide a mode, so we will need to check that from the OPENFILE statement. That is the resolver's job, so we won't do that here.

https://github.com/nyjc-computing/pseudo/blob/4a96955a465277afcb5fbe18d0a57d0e9f91a835/parser.py#L422-L434
WRITEFILE statements take in a *string* after the file identifier, and we pass that as teh `'data'` key in the `'file'` statement. WRITEFILE statements do not provide a mode, so we will need to check that from the OPENFILE statement. That is the resolver's job, so we won't do that here.

https://github.com/nyjc-computing/pseudo/blob/4a96955a465277afcb5fbe18d0a57d0e9f91a835/parser.py#L436-L446
The simplest of the four statements. Not much to say here!

ngjunsiang:
# Verifying *file statements

Thanks to our earlier work, we only need one verifier here, and one matching rule:
https://github.com/nyjc-computing/pseudo/blob/ae25adb413f79f911c71480ab6ef830cc561b75a/resolver.py#L200-L201
https://github.com/nyjc-computing/pseudo/blob/ae25adb413f79f911c71480ab6ef830cc561b75a/resolver.py#L225-L226

ngjunsiang:
The resolver stops short of actually accessing or writing to the file; that's the interpreter's job. Here, we primarily check that the user is not trying to read/write/close an unopened file, and we check that read/write operations are being requested from files that have been opened in the appropriate modes:

https://github.com/nyjc-computing/pseudo/blob/fefa2ac207724bb1abfbc769dba5388c21398692/resolver.py#L200-L223

ngjunsiang:
# Executing *file statements

Our early decision to use one `'file'` `stmt` continues to reap benefits, requiring only one rule and one executor:
https://github.com/nyjc-computing/pseudo/blob/144772bec77cf87646cee2eecf39b03234276300/interpreter.py#L125-L126
https://github.com/nyjc-computing/pseudo/blob/144772bec77cf87646cee2eecf39b03234276300/interpreter.py#L97-L98

ngjunsiang:
The resolver has done existence and mode checks, so we can skip the `name in frame` checks, and assume the mode is generally correct. So in our executor, we are mainly carrying out IO operations, and catching any possible errors.

https://github.com/nyjc-computing/pseudo/blob/d29efd553465f4863316faeb4e93ff293d7e3eac/interpreter.py#L99-L106
We finally get round to getting a file handle from Python, and storing it in the frame slot under the `'value'` key. We get a bit lazy and use the first letter of the mode (`READ, 'WRITE`, or `APPEND`) as the `open()` mode; let's see if this comes back to bite us.

https://github.com/nyjc-computing/pseudo/blob/d19a55038b9cdcbfec4e2ad779e464815c9dd79e/interpreter.py#L107-L111
We keep the file handle open precisely because we don't want to implement file cursors from scratch; here we let Python help us remember the position of the pointer (in the file handle). We call the file handle's `readline()` for each line, stripping the line break as a convenience.

https://github.com/nyjc-computing/pseudo/blob/d19a55038b9cdcbfec4e2ad779e464815c9dd79e/interpreter.py#L112-L118
To write data, we get the existing file handle, and use its `write()` method. Pseudocode requires moving the pointer to the next line after each WRITEFILE, so we check for a terminal line break and append one if necessary.

https://github.com/nyjc-computing/pseudo/blob/d19a55038b9cdcbfec4e2ad779e464815c9dd79e/interpreter.py#L119-L122
Lastly, when CLOSEFILE is called, we close the file handle and remove the file identifier from the frame.

ngjunsiang:
# Bugfixes

An incorrect ENDWHILE match: [36fc2608429e28fc227f9dd024373dcfbac2b0a3]

Incorrect extraction of file open mode: [9af8094ba99e506d8c9bb542272c699e085beb01]

Forgetting to insert frames during resolving: [a7a84e99f2d9131f08d2b954540f66cc5d9075b1], [ffc7a0922336d7447be9b687d6b0c1ae57b6875a], [21ba6a104b1aed85e47019467b4d00b133fd644a]

And pseudocode bugs:
- forgetting to add ENDWHILE [324de84b8d51357fa5ae6c72776d2d31bcfae766]
- forgetting to initialise the `Max` value [448c72fdf5376f3115772441aefdd931bcd2986a]

The latter bug points to a missing feature: because the resolver does not actually execute operations, it cannot know if variables have been initialised. The interpreter should have raised a `RuntimeError` upon encountering a NULL value. This will probably take another refactor to resolve, so let's add another issue (#33) and fix it some other time.

ngjunsiang:
# Testing

Test code:
```
DECLARE Number : INTEGER
DECLARE Max : INTEGER
Number <- 0
Max <- 10
OPENFILE "data.txt" FOR WRITE
WHILE Number <= Max DO
    WRITEFILE "data.txt", Number
    Number <- Number + 1
ENDWHILE
CLOSEFILE "data.txt"
```

Result (in `data.txt`):
```
0
1
2
3
4
5
6
7
8
9
10
```

PR #34, opened 2022-04-27
16a OOP: Expressions

ngjunsiang:
We are hitting the limit of what we can do for expressions using only dicts. Up to this point we've been shoehorning all kinds of expressions—`call` expressions, `get` expressions, binary expressions, and even unary expressions—into an expr` with a `'left'`, `'oper'`, and `'right'`.

Let's do it the "proper" way with object-oriented programming (OOP) so that we can easily extend our interpreter to support other kinds of expressions.

### DIFF
--- a/builtin.py
+++ b/builtin.py
@@ -7,8 +7,10 @@ class PseudoError(Exception):
         self.token = token
         if line is not None:
             self.line = line
-        else:
+        elif token is not None:
             self.line = token['line']
+        else:
+            self.line = None
         self.col = token['col']
 
     def msg(self):
--- a/builtin.py
+++ b/builtin.py
@@ -7,11 +7,11 @@ class PseudoError(Exception):
         self.token = token
         if line is not None:
             self.line = line
-        elif token is not None:
+        elif token:
             self.line = token['line']
         else:
             self.line = None
-        self.col = token['col']
+        self.col = token['col'] if token else None
 
     def msg(self):
         return self.args[0]
--- a/builtin.py
+++ b/builtin.py
@@ -91,6 +91,8 @@ KEYWORDS = [
 
 TYPES = ['INTEGER', 'STRING']
 
+NULL = object()
+
 OPERATORS = {
     '+': add,
     '-': sub,
--- a/interpreter.py
+++ b/interpreter.py
@@ -13,8 +13,8 @@ def executeStmts(frame, stmts):
 
 def assignArgsParams(frame, args, callable):
     for arg, param in zip(args, callable['params']):
-        name = evaluate(callable['frame'], param['name'])
-        callable['frame'][name]['value'] = evaluate(frame, arg)
+        name = param['name'].evaluate(callable['frame'])
+        callable['frame'][name]['value'] = arg.evaluate(frame)
 
 def evaluate(frame, expr):
     # Evaluating tokens
@@ -38,30 +38,30 @@ def evaluate(frame, expr):
 
 def execOutput(frame, stmt):
     for expr in stmt['exprs']:
-        print(str(evaluate(frame, expr)), end='')
+        print(str(expr.evaluate(frame)), end='')
     print('')  # Add \n
 
 def execInput(frame, stmt):
-    name = evaluate(frame, stmt['name'])
+    name = stmt['name'].evaluate(frame)
     frame[name]['value'] = input()
 
 def execDeclare(frame, stmt):
     pass
 
 def execAssign(frame, stmt):
-    name = evaluate(frame, stmt['name'])
-    value = evaluate(frame, stmt['expr'])
+    name = stmt['name'].evaluate(frame)
+    value = stmt['expr'].evaluate(frame)
     frame[name]['value'] = value
 
 def execCase(frame, stmt):
-    cond = evaluate(frame, stmt['cond'])
+    cond = stmt['cond'].evaluate(frame)
     if cond in stmt['stmts']:
         execute(frame, stmt['stmts'][cond])
     elif stmt['fallback']:
         execute(frame, stmt['fallback'])
 
 def execIf(frame, stmt):
-    if evaluate(frame, stmt['cond']):
+    if stmt['cond'].evaluate(frame):
         executeStmts(frame, stmt['stmts'][True])
     elif stmt['fallback']:
         executeStmts(frame, stmt['fallback'])
@@ -69,12 +69,12 @@ def execIf(frame, stmt):
 def execWhile(frame, stmt):
     if stmt['init']:
         execute(frame, stmt['init'])
-    while evaluate(frame, stmt['cond']) is True:
+    while stmt['cond'].evaluate(frame) is True:
         executeStmts(frame, stmt['stmts'])
 
 def execRepeat(frame, stmt):
     executeStmts(frame, stmt['stmts'])
-    while evaluate(frame, stmt['cond']) is False:
+    while stmt['cond'].evaluate(frame) is False:
         executeStmts(frame, stmt['stmts'])
 
 def execProcedure(frame, stmt):
@@ -84,7 +84,7 @@ def execFunction(frame, stmt):
     pass
 
 def execCall(frame, stmt):
-    proc = evaluate(frame, stmt['name'])
+    proc = stmt['name'].evaluate(frame)
     assignArgsParams(frame, stmt['args'], proc)
     return executeStmts(frame, proc['stmts'])
 
@@ -92,10 +92,10 @@ def execReturn(local, stmt):
     # This will typically be execute()ed within
     # evaluate() in a function call, so frame is expected
     # to be local
-    return evaluate(local, stmt['expr'])
+    return stmt['expr'].evaluate(local)
 
 def execFile(frame, stmt):
-    name = evaluate(frame, stmt['name'])
+    name = stmt['name'].evaluate(frame)
     if stmt['action'] == 'open':
         mode = stmt['mode']['word']
         assert mode  # Internal check
@@ -106,12 +106,12 @@ def execFile(frame, stmt):
         frame[name] = file
     elif stmt['action'] == 'read':
         file = frame[name]
-        varname = evaluate(frame, stmt['data'])
+        varname = stmt['data'].evaluate(frame)
         line = file['value'].readline().rstrip()
         frame[varname]['value'] = line
     elif stmt['action'] == 'write':
         file = frame[name]
-        writedata = str(evaluate(frame, stmt['data']))
+        writedata = str(stmt['data'].evaluate(frame))
         # Move pointer to next line after writing
         if not writedata.endswith('\n'):
             writedata += '\n'
--- a/lang.py
+++ b/lang.py
@@ -1,0 +1,2 @@
+class Expr:
+    pass
--- a/lang.py
+++ b/lang.py
@@ -4,7 +4,7 @@ from builtin import add, sub, mul, div
 
 
 class Expr:
-    def resolve(self):
+    def resolve(self, frame=None):
         raise NotImplementedError
 
     def evaluate(self):
@@ -51,8 +51,8 @@ class Unary(Expr):
         self.oper = oper
         self.right = right
 
-    def resolve(self, frame=None):
-        return self.right.resolve()
+    def resolve(self, frame):
+        return self.right.resolve(frame)
 
     def evaluate(self):
         return self.oper(self.right.value)
--- a/lang.py
+++ b/lang.py
@@ -8,7 +8,7 @@ class Expr:
     def resolve(self, frame=None):
         raise NotImplementedError
 
-    def evaluate(self):
+    def evaluate(self, frame=None):
         raise NotImplementedError
 
     def __repr__(self):
@@ -28,7 +28,7 @@ class Literal(Expr):
     def resolve(self, frame=None):
         return self.type
 
-    def evaluate(self):
+    def evaluate(self, frame=None):
         return self.value
 
 
@@ -41,7 +41,7 @@ class Name(Expr):
     def resolve(self, frame=None):
         return self.name
 
-    def evaluate(self):
+    def evaluate(self, frame=None):
         return self.name
 
 
@@ -55,8 +55,8 @@ class Unary(Expr):
     def resolve(self, frame):
         return self.right.resolve(frame)
 
-    def evaluate(self):
-        right = self.right.evaluate()
+    def evaluate(self, frame):
+        right = self.right.evaluate(frame)
         return self.oper(right)
 
 
@@ -76,9 +76,9 @@ class Binary(Expr):
         elif self.oper in (add, sub, mul, div):
             return 'INTEGER'
 
-    def evaluate(self):
-        left = self.left.evaluate()
-        right = self.right.evaluate()
+    def evaluate(self, frame):
+        left = self.left.evaluate(frame)
+        right = self.right.evaluate(frame)
         return self.oper(left, right)
 
 
@@ -96,8 +96,8 @@ class Get(Expr):
         slot = self.frame[name]
         return slot['type']
 
-    def evaluate(self):
-        name = self.name.evaluate()
+    def evaluate(self, frame):
+        name = self.name.evaluate(frame)
         slot = self.frame[name]
         return slot['value']
 
@@ -113,5 +113,5 @@ class Call(Expr):
         self.callable.resolve(frame)
         return self.callable.resolve()
 
-    def evaluate(self):
-        return self.callable.evaluate()
+    def evaluate(self, frame):
+        return self.callable.evaluate(frame)
--- a/lang.py
+++ b/lang.py
@@ -12,7 +12,7 @@ class Expr:
 
     def __repr__(self):
         attrstr = ", ".join([
-            getattr(self, attr) for attr in self.__slots__
+            repr(getattr(self, attr)) for attr in self.__slots__
         ])
         return f'{type(self).__class__}({attrstr})'
 
--- a/lang.py
+++ b/lang.py
@@ -103,6 +103,22 @@ class Get(Expr):
 
 
 
+def executeStmts(frame, stmts):
+    for stmt in stmts:
+        returnval = execute(frame, stmt)
+        if returnval:
+                return returnval
+
+def assignArgsParams(frame, args, callable):
+    for arg, param in zip(args, callable['params']):
+        name = param['name'].evaluate(callable['frame'])
+        callable['frame'][name]['value'] = arg.evaluate(frame)
+
+def execCall(frame, stmt):
+    proc = stmt['name'].evaluate(frame)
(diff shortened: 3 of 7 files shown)


ngjunsiang:
Taking a moment to think about all the things we need to do with an `Expr` class:
- `resolve()` to get a type
- `evaluate()` to get a value

Actually ... that's pretty much it! `'left'`, `'right'`, and `'oper'` were standardisations we had to impose to make expressions more predictable. Ultimately, though, we should only need to use an object's methods to do/get whatever we need.

So what kinds of expressions are we going to need?
- unary expressions (one operator, one operand)
- binary expressions (one operator, two operands)
- get expressions (a frame, a name)
- call expressions(a function/procedure, argument values)

Let's get started.

ngjunsiang:
https://github.com/nyjc-computing/pseudo/blob/0ab95e42597a6ede5e6d88de4ef39be5f8885725/lang.py#L1-L6
To ensure subclasses implement `resolve()` and `evaluate()` methods, we `raise NotImplementedError` from the superclass if they are not implemented by the subclass.

https://github.com/nyjc-computing/pseudo/blob/0ab95e42597a6ede5e6d88de4ef39be5f8885725/lang.py#L10-L35
We initialise the subclasses with attributes.

ngjunsiang:
Adding `repr`s for debugging: [2837078405fd95b5669d729a1afa2879bf0f389e]

Okay that's just messy. Let's abstract it into the parent class:
Add `__slot__`s: [5c031794a313a7b85e46928b235b7e122bb8f0e4]
And then use slots to generate a `repr` from the superclass: [cd908c90e045285eda07263c191f62805c40fa2a]

ngjunsiang:
# Literals

https://github.com/nyjc-computing/pseudo/blob/4047fb68ef5cf37ac92f2d63f05085ab4916fb2f/lang.py#L19-L23
Now we can figure out the `Literal` expr subclass.

https://github.com/nyjc-computing/pseudo/blob/f6548d8a91d2a48a424dd6d3a0e827a497c7e30a/lang.py#L25-L29
The simplest possible `resolve()` and `evaluate()`: a Literal is what we use when the user writes a `1` or `"Hello"` in the pseudocode. It holds an actual, determined value with a known type.

ngjunsiang:
# Unarys

https://github.com/nyjc-computing/pseudo/blob/5443c770e4c85c78805745fecf26e21c72847e48/lang.py#L33-L43
Unarys consist of an `oper` and a `right`. We are assuming here that the `oper` does not change `right`'s type. So we delegate `resolve()` to `right` by calling its `evaluate()` method. And we `evaluate()` a Unary but passing the `right` value to the `oper` (which now takes only a single value).

ngjunsiang:
# Binarys

https://github.com/nyjc-computing/pseudo/blob/599b5ff77b272438057e72b5dc151a45e9ad615f/lang.py#L47-L61
Taking a page from what we wrote in `evaluate()`, we check the `oper` to determine the return type. This will get more complicated when we support REAL types, but that's not happening yet.

Evaluation is much the same as for Unarys. Here I am avoiding complicating the `evaluate()` method; any required type-checking should be done before `evaluate()`.

We'll handle frame insertion later.

ngjunsiang:
# Gets

https://github.com/nyjc-computing/pseudo/blob/d26f41d9a8ac8ff08955680d2c26b01f9cadab82/lang.py#L65-L79
Keeping it simple, `resolve()` and `evaluate()` look very similar: both are retrieving a slot from the frame. But we won't be receiving a raw string for `name`; it is likely to be a name token, which reminds me: We are going to need a `Name` subclass for Expr as well. So we `evaluate()` `name` to get a key for `frame`.

## Name Expr

https://github.com/nyjc-computing/pseudo/blob/98be282229e11c3eb83c0def2cdd19d65aa92e9c/lang.py#L33-L42
Let's make a `Name` expr too, which `resolve()`s to a new type, `'NAME'`. And it `evaluate()`s to its name, a raw string.

ngjunsiang:
# Calls

https://github.com/nyjc-computing/pseudo/blob/0cbc0094dce4510520fcc44906278d4a18205fd6/lang.py#L96-L109
Lastly, Calls. This takes some explaining.

We are expecting `self.callable` to be a `Get` `expr`, since this is going to be scanned as a name token in the scanner. The actual callable will be stored in a slot in the `frame`. So we `evaluate()` this `Get` to access the slot, and retrieve the type from it (for `resolve()`).

As for `evaluate()` ... let's defer that implementation till later :stuck_out_tongue:

ngjunsiang:
# Parsing tokens, making Exprs

Now we refactor our parser to make `Expr` tokens instead of dict tokens.

ngjunsiang:
It's a little daunting just changing all the `makeExpr()`s into class calls 😰 and what if I change my mind and decide to use some other way to represent expr`s ...?

Let's start with splitting off `makeExpr()` into six separate functions, each of which return an appropriate `Expr` object:
https://github.com/nyjc-computing/pseudo/blob/63587e766c13b039b1bfcf136e1fffea8dac3a63/parser.py#L23-L39

Looks excessive, but I plan to merge them back into 1 function later, once I figure out a pattern.

ngjunsiang:
A nice advantage of this is it lets us replace hacky `makeToken()`s, which have no real business in a parser really:
https://github.com/nyjc-computing/pseudo/blob/707c72220565b595e2ad453c9b6da5924a3bb555/parser.py#L290-L296

With a simpler expression-making function (the [design pattern](https://refactoring.guru/design-patterns) term for this is **factory function**):
https://github.com/nyjc-computing/pseudo/blob/ac5f17da608a6f890923e76bb1e5f3d7b0f434d9/parser.py#L290-L291

ngjunsiang:
## Refactoring

Let's start by replacing the positional arguments in our `make*Expr()` function calls with keyword arguments: [6f853a8559f8e2a5415e15132a8c90b3c4aebb82]

I am aiming to replace `makeLiteralExpr()`, `makeNameExpr()`, etc with a new `makeExpr()` factory that *determines which Expr to return* based on the provided keyword arguments:
https://github.com/nyjc-computing/pseudo/blob/e0a595610593c51ae6eb8e5d7e56880dd5322e8c/parser.py#L23-L30

The `*,` indicates that positional arguments are not accepted; keyword arguments only!

It does so by checking the provided keyword arguments:
https://github.com/nyjc-computing/pseudo/blob/16f8bef1a76a1f859110a6dc2f9db41f25001add/parser.py#L30-L44

ngjunsiang:
Since `makeExpr()` can now perform the job of any of the `make*Expr()` functions with the exact same keyword arguments, we replace all those calls with `makeExpr()` calls: [e98972071abf3df149a6e5fab6dd46ab70ff9e81]

ngjunsiang:
A whole bunch of bugfixes before testing:

b0ccd781f4a21f3d660a0bed49a2daa261bf81f6
6faff09affd2d39bf2f314c1189760242871f9e8
13ea9359428754db49bcf21ddd2f6093882b092b

ngjunsiang:
And more minor bugfixes in our `Expr` class:
[84efbd334267abe3d1a6f04dc4b151743cbc72d4], [98d7d77bb071d2970008e01debf2fd722e2ea235]

and in the parser:
[d542dba40f917deef4d11c43028319e25e1c5c73], [7e8e13fb3938cedea81056aa4cc5a8792d87f6b2], [f49fd954293e90a8d8659cd0c5068bcad3cb6faa]

And now our parser produces `Expr`s in place of expr dicts:
```
{'rule': 'declare', 'name': Name('Number'), 'type': {'line': 1, 'col': 17, 'type': 'name', 'word': 'INTEGER', 'value': None}}
{'rule': 'declare', 'name': Name('Max'), 'type': {'line': 2, 'col': 14, 'type': 'name', 'word': 'INTEGER', 'value': None}}
{'rule': 'assign', 'name': Name('Number'), 'expr': Literal('integer', 0)}
{'rule': 'assign', 'name': Name('Max'), 'expr': Literal('integer', 10)}
{'rule': 'file', 'action': 'open', 'name': Literal('string', 'data.txt'), 'mode': {'line': 5, 'col': 24, 'type': 'keyword', 'word': 'WRITE', 'value': None}, 'data': None}
{'rule': 'while', 'init': None, 'cond': Binary(Name(Name('Number')), <function lte at 0x7f883a501040>, Name(Name('Max'))), 'stmts': [{'rule': 'file', 'action': 'write', 'name': Literal('string', 'data.txt'), 'mode': None, 'data': Name(Name('Number'))}, {'rule': 'assign', 'name': Name('Number'), 'expr': Binary(Name(Name('Number')), <function add at 0x7f883a4f9b80>, Literal('integer', 1))}]}
{'rule': 'file', 'action': 'close', 'name': Literal('string', 'data.txt'), 'mode': None, 'data': None}
```

ngjunsiang:
## Working around frame insertion

This leaves us with a very annoying bug. In `makeExpr()`, we do an explicit `None` check to decide what kind of `Expr` to return:
https://github.com/nyjc-computing/pseudo/blob/98d7d77bb071d2970008e01debf2fd722e2ea235/parser.py#L30-L34

But in `value()`, we insert a `None` in place of `frame` for later insertion by the resolver:
https://github.com/nyjc-computing/pseudo/blob/98d7d77bb071d2970008e01debf2fd722e2ea235/parser.py#L83-L90

This is why we have a `Name(Name('Number'))` in our `while` stmt above, where we should be expecting a `Get(Name('Number'))`!

Resolving this is tricky: replacing `frame` with anything other than `None` means we need to use some explicit checking for a string. Instead, let's finally make our own None type. Coincidentally, 9608 pseudocode does define its own `None`-equivalent, called NULL.

ngjunsiang:
## NULL value

We create our own NULL value by calling Python's `object()`, which gives us a unique object with a unique object id:
https://github.com/nyjc-computing/pseudo/blob/9f709e858f5ab61905851f8f545ad21bdbdb50be/builtin.py#L94

We can use an `is NULL` check to see if something has a value of NULL.

Import it into our parser:
https://github.com/nyjc-computing/pseudo/blob/97a8f36f5457179386fbaa22b44aeacc23f29091/parser.py#L1

And use it in place of `None` for our frame:
https://github.com/nyjc-computing/pseudo/blob/97a8f36f5457179386fbaa22b44aeacc23f29091/parser.py#L83-L90

Now we have a `Get(<frame>, Name('Number'))` expression for our `while` `cond`:
```
{'rule': 'while', 'init': None, 'cond': Binary(Get(<object object at 0x7f3ebcde4e80>, Name('Number')), <function lte at 0x7f3ebcc1e040>, Get(<object object at 0x7f3ebcde4e80>, Name('Max'))), 'stmts': [{'rule': 'file', 'action': 'write', 'name': Literal('string', 'data.txt'), 'mode': None, 'data': Get(<object object at 0x7f3ebcde4e80>, Name('Number'))}, {'rule': 'assign', 'name': Name('Number'), 'expr': Binary(Get(<object object at 0x7f3ebcde4e80>, Name('Number')), <function add at 0x7f3ebcc16b80>, Literal('integer', 1))}]}
```

Our NULL object is looking pretty ugly there, but fortunately our users (pseudocode programmers) will never get to see it.

ngjunsiang:
# Resolving Exprs

Finally we move on to the resolver. Our first hiccup: resolving involves frame insertion. That means the `resolve()` method of our `Expr`s must take in the `frame` as an optional argument: 6d2a54f45b17f3b01025833db36557ec8631d3af]

`Literal` and `Name` don't need a frame and therefore no changes there.

`Unary` must resolve its `right`, in case it's a Get` Expr (e.g. `-Var`):
https://github.com/nyjc-computing/pseudo/blob/5008eb80237b454f7f58313263423db7f9babbe8/lang.py#L54-L55

`Binary` must resolve its `left` and `right`:
https://github.com/nyjc-computing/pseudo/blob/4af23ce87b76d3b73d68d207be896702a07b836a/lang.py#L69-L75

`Get` must replace its `frame` with the provided one:
https://github.com/nyjc-computing/pseudo/blob/c11ee8f5b304fdd079e6d9a48ba21ec90189b72e/lang.py#L89-L94

`Call` must resolve its `callable`, which is not the procedure/function itself but a `Get` Expr:
https://github.com/nyjc-computing/pseudo/blob/bb74075b1e40908ea4483c8fb57522bd4aaf387e/lang.py#L109-L111

This feels weird: why not `evaluate()` the `Get` Expr and replace the `callable` attribute? Because this lets us use `Get.resolve()` to retrieve the type, and `Get.evaluate()` to retrieve the callable itself!

ngjunsiang:
After working out the kinks in `lang.py`, let's move on to `resolver.py`.

Having `Expr.resolve()` methods means we shouldn't need `resolve()` anymore, right? Let's try to eliminate all `resolve()` calls and replace them with equivalent `Expr.resolve()` methods.

INPUT:
https://github.com/nyjc-computing/pseudo/blob/1af3e40e02074d195a1091bc3154e4ed4d76f9e5/resolver.py#L78-L84

We'll also need to fix our `Name.resolve()` to make it return the word. Yes, that's exactly the same as `evaluate()`; more things we will have to rethink and refactor later 😩
https://github.com/nyjc-computing/pseudo/blob/b1ab2d9fa121f429dfd73a095ad7a42fb397d37c/lang.py#L41-L42

DECLARE:
https://github.com/nyjc-computing/pseudo/blob/eb366cff74d1961aa8cf413026ac6055ff66a56f/resolver.py#L86-L89

Feels like a step backwards; we are back to grabbing keys from dicts manually. Sometimes we gotta take a step back before we can take a step forwards; we have to do this because our `stmt`s are still dicts and not their own object type yet. That will come later.

Minor fix: forgot to pass `frame` to the `resolve()` methods. [fbb89e3185c48423b87b6d0ae5e43439f2c21843]

ASSIGN:
https://github.com/nyjc-computing/pseudo/blob/b8528330f51168c598888d34041169b5013b89e9/resolver.py#L91-L93

CASE/IF/WHILE:
[d56510c65d6b56586b6078d127e7f735dd4f1efe]
These statements rely on a conditional Expr, which we now resolve using the Expr's `resolve()` method.

PROCEDURE/FUNCTION:
[6e67dcae6dfc31dfa9f252140d04f4eab65efdac]
The changes here primarily concern name-resolving. You can see how it starts to get a bit mindless: having a polymorphic `resolve()` method makes things much easier!

CALL:
[071446278cb9ccb29b8d1fbfea606982cb80b5c4]
Now our work starts to reveal its value: instead of `resolve()` having to work differently for different `expr`s, now we have each `Expr` decide what it should do to `resolve()` itself. We've moved the complexity out of `resolve()`, into the `Expr.resolve()` methods where it's slightly more manageable.

RETURN:
[fe8ceab03294c47f529ae2100c86557f686bfab3]

*FILE:
[a00fe50f6144ff4a7772a91b8794abcb462fc7ab]

And lastly in our helper functions:
[e92000d0460ad5dab6888b01b0f283f95ebcb4ba]

ngjunsiang:
# Testing

The moment of truth: ~~deleting~~ commenting out `resolve()` and see what happens:
```
  File "lang.py", line 42, in resolve
    return self.name['word']
TypeError: string indices must be integers
```
Oops, names don't resolve properly.

```
> lang.py(42)resolve()
-> return self.name['word']
(Pdb) self.name
'Number'
```
Turns out instead of passing a token, we passed the word directly to `Name`. That's a good idea; our resolver should not have to concern itself with tokens already. We'll do another refactoring check for that later.

The fix:
https://github.com/nyjc-computing/pseudo/blob/18533535af4278c97335505cf06ac2541629ce8a/lang.py#L41-L45

`expectTypeElseError()` is in limbo now; it isn't fully refactored to handle `Expr`s, so we insert a temporary workaround:
https://github.com/nyjc-computing/pseudo/blob/18533535af4278c97335505cf06ac2541629ce8a/resolver.py#L10-L24

This means we might sometimes raise errors without a token, so we also add another workaround for this: [a81c9b0242334d5115e5be166ced9a979d8facfa], [0cf5095849cfc50f9d6a05d99d43e4fab6372790]

(We'll get round to fixing this, once `Expr`s work!)

ngjunsiang:
```
None: Expected 'INTEGER', got 'integer'
```

Oops, we need to standardise our cases: [80787d159b41a7383c6538751adfe7ded5d2d997], [42ba1f632e89a66b633edefcc47c92092c5ee4f0]

And our resolver works!

ngjunsiang:
# Evaluating Exprs

And it's the same story for `evaluate()`! Let's try to eliminate it, replacing it with `Expr.evaluate()`.

First, we turn our attention to the `Expr.evaluate()` methods and correct them for recursion: [184eae4c1d1f3208812540c358d68dabede5a35c]

`Call` presents a thorny edge case: Should we `evaluate()` the callable here, or elsewhere? Doing it in `lang.py` may involve pulling in code from the other modules, and we want to keep `lang.py` as dependency-free as possible. For now, let's just return the callable; it is going to be a `Get` expr so we `evaluate()` that to get and return the callable itself.
https://github.com/nyjc-computing/pseudo/blob/eadf6827566620a29631be78c20f49debfb9bc3e/lang.py#L116-L117

ngjunsiang:
Now we do a refactor, converting all `evaluate()` function calls to `Expr.evaluate()` method calls. It's quite a straightforward conversion so I won't do a blow-by-blow rundown: [01e97e606595ed80d8c0a79fb7bd87dccd0e7b7d]

And remembering to pass the frames to `Expr.evaluate()` in `lang.py`: [855ea90afc67bcab8e06c886c6e645454da94279]

And we're golden: it works!

ngjunsiang:
# That tricky Call ...

Now let's turn our attention back to `Call` Expr:
https://github.com/nyjc-computing/pseudo/blob/855ea90afc67bcab8e06c886c6e645454da94279/interpreter.py#L30-L34

`evaluate()` calls `execCall()`:
https://github.com/nyjc-computing/pseudo/blob/855ea90afc67bcab8e06c886c6e645454da94279/interpreter.py#L86-L89

and `execCall()` in turn depends on `assignArgsParams` and `executeStmts()`:
https://github.com/nyjc-computing/pseudo/blob/855ea90afc67bcab8e06c886c6e645454da94279/interpreter.py#L8-L17

Let's put this code into `lang.py` and see what we can simplify:
https://github.com/nyjc-computing/pseudo/blob/e096fb32605292b31450d6fb775dc1c528271dfd/lang.py#L106-L134

ngjunsiang:
Unpack those functions and dump them into `Call.evaluate()`:
https://github.com/nyjc-computing/pseudo/blob/741c9cf0c28802e37700a69f8947d9c087897658/lang.py#L116-L125

😣! On line 123 ... we have an `execute()` dependency! And `execute()` is a helluva long function:
https://github.com/nyjc-computing/pseudo/blob/741c9cf0c28802e37700a69f8947d9c087897658/interpreter.py#L124-L150

Fine ... we'll temporarily import it from `interpreter.py`:
https://github.com/nyjc-computing/pseudo/blob/afb66cdfb5de7eb47fa208a81f163e96ebb45b11/lang.py#L4

And continue with encapsulating our statements using OOP as well. Next chapter!

PR #36, opened 2022-04-29
16c OOP: Expression Statements

ngjunsiang:
# Introducing Expression Statements

There are some pseudocode features that straddle the line between expressions and features. For instance:

`DECLARE Number : INTEGER`

and

`PROCEDURE(Number : INTEGER)`

In both cases, `Number : INTEGER` declares a variable `Number` of type `INTEGER`. DECLARE means this variable is in the global space, while PROCEDURE() means the variable is in the procedure's local space.

So `Number : INTEGER` is an expression, but has to be executed like a statement (since it has effects, namely declaration of a variable).

Then there's the story of CALL, which essentially does the same thing as a function call but for a procedure.

Let's make a new kind of statement for these, so we can `verify()` and `execute()` them. We'll call them ... `ExprStmt`s.

### DIFF
--- a/builtin.py
+++ b/builtin.py
@@ -66,12 +66,6 @@ def ne(x, y):
 def eq(x, y):
     return x == y
 
-def get(frame, name):
-    return frame[name]['value']
-
-def call(func, args):
-    return func, args
-
 
 
 # Token types
--- a/builtin.py
+++ b/builtin.py
@@ -5,13 +5,15 @@ class PseudoError(Exception):
     def __init__(self, msg, token, line=None):
         super().__init__(msg)
         self.token = token
+        self.line = None
+        self.col = None
         if line is not None:
             self.line = line
-        elif token:
+        elif type(token) is dict:
             self.line = token['line']
+            self.col = token['col']
         else:
             self.line = None
-        self.col = token['col'] if token else None
 
     def msg(self):
         return self.args[0]
--- a/interpreter.py
+++ b/interpreter.py
@@ -63,11 +63,11 @@ def execOutput(frame, stmt):
     print('')  # Add \n
 
 def execInput(frame, stmt):
-    name = stmt.name.accept(frame, evaluate)
+    name = stmt.name
     frame[name]['value'] = input()
 
 def execAssign(frame, stmt):
-    name = stmt.name.accept(frame, evaluate)
+    name = stmt.name
     value = stmt.expr.accept(frame, evaluate)
     frame[name]['value'] = value
 
@@ -96,7 +96,7 @@ def execRepeat(frame, stmt):
         executeStmts(frame, stmt.stmts)
 
 def execFile(frame, stmt):
-    name = stmt.name.accept(frame, evaluate)
+    name = stmt.name
     if stmt.action == 'open':
         assert stmt.mode  # Internal check
         file = {
--- a/interpreter.py
+++ b/interpreter.py
@@ -1,4 +1,3 @@
-from builtin import call
 from builtin import RuntimeError
 
 
--- a/interpreter.py
+++ b/interpreter.py
@@ -1,4 +1,5 @@
 from builtin import RuntimeError
+from lang import Literal, Declare, Unary, Binary, Get, Call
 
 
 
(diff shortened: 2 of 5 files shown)


ngjunsiang:
Behold, the ExprStmt:
https://github.com/nyjc-computing/pseudo/blob/0e6f0f26de5501649928df38e2769f0df36905d2/lang.py#L148-L152

Our Expr is probably going to need an `accept()` method to implement the Visitor pattern, I can just feel it. Let's add it first: [33b80f3d26a001126e4a7a77fcfe0da3532b278b]

ngjunsiang:
# Resolving ExprStmts

## Resolving Declare

In `resolver.py`, `resolveDeclare()` shall be a function that takes in a `frame` and Declare `expr`, and declares the represented variable in frame`:
https://github.com/nyjc-computing/pseudo/blob/dcb6de47dec4d94c4e1120f4a6a6ab025846adab/resolver.py#L46-L53

Our Visitor pattern's `accept()` is coming in handy now. Come to think of it, why do we still need `verifyDeclare()` if it only does one thing? Let's replace it with the Visitor pattern: [45a22a7b8f494f21f4ffcc95ba0d24ad11c7dd01]

And updating all our uses of `var` in `params`:
https://github.com/nyjc-computing/pseudo/blob/3c0f3268bc132536d660ac2473ef0772ef7387f4/resolver.py#L107-L113

## Resolving Call
Now we translate `verifyCall()` into `resolveCall()`. First, let's set things up the way we want.
https://github.com/nyjc-computing/pseudo/blob/12e94d4ece443a95b46f6646435533261c33572a/resolver.py#L51-L67

We use `resolveGet()` to insert the frame into a Get Expr, also adding a little `assert` as a treat.

We use `get` to retrieve a value from a frame, at the same time doing the expected existence and value checks.

And then we can use them in `resolveCall()`.

Some minor fixes not quite related to our work here: [89bc99931c5631de09464e683dc97445c96daf3a], [4dbc31df387b5c20690747a2b199d4696e403db8]

Continuing with our vein of avoiding tokens in expressions and statements, we also try to avoid unnecessary nesting. First thing to go: `Name` exprs; we can just use strings!
[b0fae35d892f238f52f44979c44b1bd624d2df0b], [44d6005885a5fc67826a0ac5d953ed4332ea5e3a], [6b0929b379d41238eb0eab067f8ccd93687515ac]

ngjunsiang:
# Bugfixes

`Literal('STRING', 'data.txt'): File not open`

Looks like not all names have been unpacked ... here our OPENFILE statement, while being resolved, is trying to open a Literal instead of a filename string. Unlike with names, we do want to pass Literals and not just strings though, because Literals contain type information as well. And we need to type-check the filename.

Time to simplify type-checking too. We'll just check the type-words directly instead of having to resolve, since we're trying to bypass tokens.
https://github.com/nyjc-computing/pseudo/blob/f642d1c7f5c36d9784cafba3aaa0f277a393a759/resolver.py#L10-L12
[]

To resolve a generic Expr, we'll need to reintroduce `resolve()`, this time round as a factory function that determines the `resolve*()` function to use depending on the kind of Expr:
https://github.com/nyjc-computing/pseudo/blob/3000e9bb1ca59e9cc89cede9faf28ca32a1d969f/resolver.py#L96-L108

And the actual resolvers:
https://github.com/nyjc-computing/pseudo/blob/3000e9bb1ca59e9cc89cede9faf28ca32a1d969f/resolver.py#L45-L70

ngjunsiang:
Finally we can eliminate our cursed Calling Stmt class: [bb7026452fc0a978927f0eb52dda0f57002ca675]

ngjunsiang:
# Interpreting Exprs

Let's make a major commitment to the Visitor pattern here, and only use the `Expr.accept()` method to do resolving and evaluating: [6365e90e30f76c682a46ce24accc293aeb8b49cc], [a955be643e0cccc2b86a1c7a195a47997f0bbb1c]

ngjunsiang:
Finally we can implement the Expr `eval*()`s: [7af33ad1326d59b72e4dd00ec46a234877514edc]

Because we've done our type- and existence-checks in the resolver, the evaluation is relatively simple! Except for Call ... which looks like this:
https://github.com/nyjc-computing/pseudo/blob/8f6083d847486c2507f35f56ee942e1ef446a2fb/interpreter.py#L34-L44

Still, without having to do type-checking, it looks much cleaner. And because we replaced the Declare Exprs with frame slots, we can directly assign values into the slots (line 39).

The complication here is with checking for, and returning, any return value that isn't `None`.

ngjunsiang:
And we have working ExprStmts! We combined Declare, Call, and Return statements, which contain an expression, into one kind of statement. And in the process simplified a lot of statement parsing.

We have one more problem to worry about in statements, which we will get to in the next chapter.

PR #37, opened 2022-05-01
16d OOP: Variables

ngjunsiang:
The first place in our code we see dicts being used is in our scanner, where tokens are represented as dicts. Let's not touch this for now; it's relatively minor and doesn't affect the actual execution of code.

The next place they are used in in the resolver, where `frame`s use them to store values, in a dict with a `'type'` key and a`'value'` key.

This looks kind of similar to Literal Exprs, which also have a type and a value. In fact, because 9608 pseudocode is statically typed, every value necessarily has a type!

# Typed values

If we are going to need containers to hold the types and values of each *thing* in pseudocode, what would we call this container? For lack of a better term, let's just call it a **TypedValue**. We can use this to represent Literals, and when they are stored in a **frame**, we call them Variables.

TypedValues are containers for type and value; they are not Exprs or Stmts, and we will not be adding an `accept()` method.

https://github.com/nyjc-computing/pseudo/blob/58d2c2840bdc9df4309e9880bcac39a038180bb7/lang.py#L1-L9


ngjunsiang:
We can, however, have Literal inherit from TypedValue since their constructors are practically identical:

https://github.com/nyjc-computing/pseudo/blob/e51db8ce86b971b001448d69892ff3172f23e193/lang.py#L42-L46

ngjunsiang:
A minor bugfix I just caught: evaluating a Get expr should return a `value` and not a slot. This is going to be changed later, so I wanted to add this change so we don't miss it: [b421440f2b2e73634bb8ca126ce17b16d8e97116]

And now we switch to TypedValue by replacing all uses of `'type'` and `'value'` in slots with the attribute equivalents: [3fc5d14f229e134d9304d94d6161a9ce0ee95254]

ngjunsiang:
Minor bugfixes after some quick testing, and we can move on to the interpreter: [3090376101b8b6a1a171b2f3cee8ce8c922685d4], [0303de480db4998cc282df8cd5b12a2c0c6a8cf8]

# Using TypedValue in interpreter

Now we do the same for the interpreter, replacing all uses of slot dicts with TypedValue: [5bdec1340f57920282b9c50d699c4bbd5c38bbfe]

ngjunsiang:
We'll need some other niceties as well. Here's a `__repr__()` for TypedValue:
https://github.com/nyjc-computing/pseudo/blob/d9649b35a5a891e76b90d70b0e969aa13e2208b4/lang.py#L11-L12

We used a `get` helper in `resolver.py`. Let's use a similar pattern to reduce the number of times we access `value` attribute of a TypedValue directly, by introducing `getValue()` and `setValue()` helpers:
https://github.com/nyjc-computing/pseudo/blob/bbdadd397cea12ee048a33d0c4cff7bd5b08bd16/interpreter.py#L15-L27

And then using them: [887d88221ac9970ba6043e938dcdb3fcd8de7893]

ngjunsiang:
# Bugfixes

A bunch of small fixes:
- Missed out a TypedValue conversion: [d394d4de6861dc69e3f5b6c495e9a3fe59c00bc6]
- Missed out one place to use the `get` helper: [223fd1f73cf161fae0f0eb83ae7565a77b12ae7f]
- A fix to resolve any Exprs in `verifyFile()`: [1abb19820acd19863b69a18eeb13b873b2f4f91a]

ngjunsiang:
With variables now made into objects, it's a bit easier to think about some things that we missed out earlier, like type-checking in file operations.

Since files can be opened/closed within procedures, static type-checking would not be sufficient to ensure that a file being read from/written to has definitely been opened beforehand, especially if this is done in a procedure with a BYVALUE parameter. Let's move it from a static check in the resolver (which raises `LogicError`) to a realtime check in the interpreter (which raises `RuntimeError`):
Adding checks that raise `RuntimeError`: [dd16b07be84f817a135925ffca96b747d2f2371d]
And removing ones that raise `LogicError`: [e11af7994c49c1e5d404787f730f076c25ebb7df]

ngjunsiang:
I'll need to refactor the helpers a little; when opening a file, I want to add a TypedValue to the frame without raising a `RuntimeError`. So we'll split the `setValue()` helper into two functions:
https://github.com/nyjc-computing/pseudo/blob/1fe61f25eac6650f3f50dd0c9e2422f7d0326097/interpreter.py#L23-L37

We also add an optional `errmsg` keyword argument for `getValue()`:
https://github.com/nyjc-computing/pseudo/blob/589ded434270d1224d647134072cc9670aad47f6/interpreter.py#L15-L21

Let's use them for reading/writing/closing operations:
https://github.com/nyjc-computing/pseudo/blob/00922d185df4fee4d515f0e3d60734af790b4d7c/interpreter.py#L133-L155

We pass `"File not open"` to our `getValue()` helper so the user isn't misled into thinking the filename is an undeclared variable.

ngjunsiang:
# Bugfixes

- Embarrassing oversight: `setValue()` can't work for nonexistent slots - [13340669fae3e28194a2167ceadae82b75ccf565]
- And an embarrassing typo in the `__repr__()` method: [c356d087542d551c1986d01f9742b902b159ef94]
- Avoiding repetition by evaluating `name` before checking file mode: [a043409e1a369e62554ba45cfb1a22e0431e70e6]
- Since `getValue()` returns us the value without type information anymore, we switch to relying on Python's file handler's `mode` attribute to check the file open mode: [38aad2cbc495f8257af39dabb017977636750e33]

ngjunsiang:
# Helpers

I really like how we used the `getValue()` and `setValue()` helpers in the interpreter. It will make OOP much easier when we get to frames. Let's do that in the resolver too.

These are the helpers we'll add and use:
- declaredElseError: [9f3abaf3f8f0e9dc0b6f0530daa8d4d2c0852d2c]
- declareVar: [84a6e804224df457498d0ab5784267632fce12e3]
- getType: [0f23a9d3d1be93d99960ec2c7c2a585405f98797]
- getValue: [69b6bc970a931693fec86e39bde01acc3dd41d5f]
- setValue: [89af1bde5f5d44bf2728beb6d614414948cba9e7]

ngjunsiang:
Let's give `interpreter.py` a similar set of helpers as well:
https://github.com/nyjc-computing/pseudo/blob/c0ccd4b560306e828a0d7da73875ae542f85e07e/interpreter.py#L9-L56

And then use them: [45fc8bea537c787c73701a97e197ee2728d1264e]

Our code is looking less and less like detailed, imperative code. There's more and more standardised, abstracted use of helper functions.

ngjunsiang:
And with that we have used OOP to create a TypedValue class.

Before we do the same to `frame`, let's talk about the kinds of values that are stored in a TypedValue as well. Next chapter.

PR #38, opened 2022-05-03
16e OOP: Values

ngjunsiang:
This is a short chapter on the kinds of values that 9608 pseudocode deals with.

We store the following simple data types in `frame`, as their corresponding Python types:
- STRING (as Python `str`)
- INTEGER (as Python `int`)

Besides the above, we also have:
- FUNCTION
- PROCEDURE
- FILE

At this point we are representing them as dicts. Let's finish up the OOP job by turning them into Values.

### DIFF
--- a/builtin.py
+++ b/builtin.py
@@ -87,7 +87,7 @@ KEYWORDS = [
     'READFILE', 'WRITEFILE', 'CLOSEFILE',
 ]
 
-TYPES = ['INTEGER', 'STRING']
+TYPES = ['INTEGER', 'STRING', 'FILE']
 
 NULL = object()
 
--- a/interpreter.py
+++ b/interpreter.py
@@ -6,7 +6,7 @@ from lang import Literal, Unary, Binary, Get, Call
 
 # Helper functions
 
-def expectModeElseError(exprmode, expected, errmsg="Expected", name=None):
+def expectTypeElseError(exprmode, expected, errmsg="Expected", name=None):
     if type(expected) is str:
         expected = (expected,)
     if not exprmode in expected:
@@ -151,7 +151,7 @@ def execFile(frame, stmt):
     elif stmt.action == 'read':
         file = getValue(frame, name, "File not open")
         mode = getType(frame, name)
-        expectModeElseError(mode, 'READ')
+        expectTypeElseError(mode, 'READ')
         varname = stmt.data.accept(frame, evaluate)
         # TODO: Catch and handle Python file io errors
         line = file.readline().rstrip()
@@ -160,7 +160,7 @@ def execFile(frame, stmt):
     elif stmt.action == 'write':
         file = getValue(frame, name, "File not open")
         mode = getType(frame, name)
-        expectModeElseError(mode, ('WRITE', 'APPEND'))
+        expectTypeElseError(mode, ('WRITE', 'APPEND'))
         writedata = str(stmt.data.accept(frame, evaluate))
         # Move pointer to next line after writing
         if not writedata.endswith('\n'):
--- a/interpreter.py
+++ b/interpreter.py
@@ -7,7 +7,7 @@ from lang import Literal, Unary, Binary, Get, Call
 # Helper functions
 
 def expectModeElseError(exprmode, expected, errmsg="Expected", name=None):
-    if not type(expected) is str:
+    if type(expected) is str:
         expected = (expected,)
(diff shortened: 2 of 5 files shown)


ngjunsiang:
# Value

Starting with the base class:
https://github.com/nyjc-computing/pseudo/blob/4f50cd52b574889c0946fbf9752cc8311ecb15dc/lang.py#L1-L5

Let's make an intermediate parent class for Function and Procedure since they are so similar:
https://github.com/nyjc-computing/pseudo/blob/93cc6dfeed7cc9aa9264f66afd8b54b4af81ce8a/lang.py#L9-L26

And then our work for Function and Procedure becomes much easier:
https://github.com/nyjc-computing/pseudo/blob/2a1ab1498c81eb1dab5a48e175a5e924ac68d637/lang.py#L30-L31
https://github.com/nyjc-computing/pseudo/blob/2a1ab1498c81eb1dab5a48e175a5e924ac68d637/lang.py#L35-L36

ngjunsiang:
Oops; we already have a Callable which subclasses Stmt. Since this isn't actually a callable, just a statement that declares one, let's rename it to ... hmm, I don't know ... ProcFunc? Yeah let's go with that: [c019702ff936e6ac0bb16678bcaba4c5c4617203]

Then we use them in the resolver: [992086b3bbace286fe8bab7b7d6baf82437c02af]
I did a quick search and realised we don't use the `passby` key at all in our interpreter. We already handled that in the resolver, by passing the slot from the global frame for BYREF, or by creating a new slot for BYVALUE. So I didn't include them in the Function and Procedure classes.

ngjunsiang:
Here's our chance to fix a bug in our current code: we don't separate FUNCTION calls (using `()`) and PROCEDURE calls (using CALL) - #39

Now that we have Functions and Procedures as separate objects, we can add helpers to differentiate them:
https://github.com/nyjc-computing/pseudo/blob/6016811997642145d1afec756e7b38d008e7a26b/resolver.py#L12-L16

Instead of using a single `resolveCall()`, let's create two entry points, one for functions and one for procedures:
https://github.com/nyjc-computing/pseudo/blob/1eae28f2ca81c3c48836eeb84947fc0339e48024/resolver.py#L96-L108

They both do their own frame insertion and type-checking before calling `resolveCall()`, which is now really only responsible for resolving the actual call:
https://github.com/nyjc-computing/pseudo/blob/0d75297c113817566cd643c54249e8327f6ee9e9/resolver.py#L110-L129

And we use the entry points in place of our previous `resolveCall()`s:
https://github.com/nyjc-computing/pseudo/blob/0d75297c113817566cd643c54249e8327f6ee9e9/resolver.py#L142-L143
https://github.com/nyjc-computing/pseudo/blob/0d75297c113817566cd643c54249e8327f6ee9e9/resolver.py#L254-L255

ngjunsiang:
Wrapping up with some niceties, like a `repr`: [491ff6ee5c21405522e36cc4e32d93fd983ce6cb], [1c67f6a1bdcb80a985d30d78ffb4d092c7b58485]

ngjunsiang:
# Files

Okay, on to files. Right now they are put in the frame as raw Python file handles, which I'm ... not too comfortable with :grimacing:. It's also that files are stored with their mode as the type, instead of something more intuitive like 'FILE'.

Let's fix that.

We start by adding a new type:
https://github.com/nyjc-computing/pseudo/blob/8663b6111cace0e37149a9365f3998183d50b0ed/builtin.py#L90

And a new type of Value:
https://github.com/nyjc-computing/pseudo/blob/8663b6111cace0e37149a9365f3998183d50b0ed/lang.py#L47-L67

Oops, the File name is already used by the File stmt. Let's rename that to FileAction:
[d27dae57c72c89d28f46abd6ab4349a9268fa401]

Let's make some Files when we open a file:
https://github.com/nyjc-computing/pseudo/blob/6ffc4b478ebe01c07bd359a88c47d9512de21bb0/interpreter.py#L147-L151

Now we can type-check for FILE in addition to its open mode: [35d3efb6bff6622f357e298b90aaf337823f8eda]

And remembering to use its attributes: [ae88973b1bc2b15c8af7158f987ebfe9d7b53e5c]

ngjunsiang:
And we are done! We have the infrastructure for adding new kinds of values later, including ARRAY, custom structs, etc.

ngjunsiang:
Fixes #39

PR #40, opened 2022-05-03
16f OOP: Frames

ngjunsiang:
Finally, we're down to the last dict ripe for conversion: the `frame`.

There's nothing wrong with a dict for a frame; what's going on is I'm tired of repeating helper functions for dealing with frames. These helper functions work primarily with frames and names, and it makes good sense to bundle them together so that we can use those functions wherever the frame goes.

Let's start with a Frame class:
https://github.com/nyjc-computing/pseudo/blob/c1ffed90f1bd74b3e3104c97d247b46cb5eca421/lang.py#L86-L133

### DIFF
--- a/interpreter.py
+++ b/interpreter.py
@@ -21,38 +21,6 @@ def undeclaredElseError(frame, name, errmsg="Already declared", declaredType=Non
     if frame.has(name):
         raise RuntimeError(errmsg, name)
 
-def declareVar(frame, name, type, errmsg="Already declared"):
-    """Declare a name in a frame"""
-    undeclaredElseError(frame, name, errmsg)
-    frame.declare(name, type)
-
-def getType(frame, name):
-    declaredElseError(frame, name)
-    return frame.getType(name)
-
-def getValue(frame, name, errmsg="Undeclared"):
-    """Retrieve value from frame using a name"""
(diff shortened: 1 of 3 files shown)


ngjunsiang:
Lets replace all the places where we treated frame as a dict, reworking them into method calls: [ac30c99cc37384a6eabfc3cab3d9f84ab8a1076a], [8d390402f5c314a2e316d9a30a01bb7e29757b83], [f17620215ca47198c0ca6543a9034ffcebdf7c7a]

ngjunsiang:
Yep, we are trying to eliminate all uses of the helper functions for frames; we have replaced them with frame methods.

Since the frame methods don't do type-checking, we still have to keep the error methods, and use them where appropriate. Still, that means we've managed to replace two sets of helper functions (in the resolver and interpreter) with one set of methods (in Frame). Overall, it's still a win for maintainability.

ngjunsiang:
And now we can remove the helpers: [69abe4003612e3ad43e2acda22564b8227943fce], [626a3433a851eb4f86ce72eb52622f171c1643a3]

ngjunsiang:
A couple quick bugfixes: [03b2ba53d079b2891b407c0c8b7cd840dedd0ee9]

PR #43, opened 2022-05-04
16h OOP: Tokens

ngjunsiang:
Finally, let's objectify tokens, making a Token class. Our code works fine as it is ... but why work with dicts if an object makes things clearer?


ngjunsiang:
Start with a Token class:
https://github.com/nyjc-computing/pseudo/blob/31f3928df82253d70042e85cfc5d0f674ad97c0b/lang.py#L1-L14

Replace all our `makeToken()` calls with Token instantiations: [1c31405a3a5a24a1f84dbfe0cb4d179c36a04da8], [f40c1c70341314e6d29a82e8903dcb8532363c8a]

And update all uses of token to convert keys to attributes: [9d0c9da3638c4d5ed26d432f3dec6a56a0ad5b4c]

## Cleanup

Now we can remove unnecessary imports, breaking the cursed dependency of our parser on the scanner: [42e10dcf19300276fe522cfff830f69a2f6509eb]

And eliminate the Return Stmt, which should have been done a few chapters ago: [0562ab9273cc59347b98552a6d5aa2e0d5cc0c8c]

ngjunsiang:
And we have Token objects!

PR #44, opened 2022-05-04
17 Statement hierarchies

ngjunsiang:
We almost have all the infrastructure in place for our interpreter. One last little niggle: Statement hierarchy.

Looking at the structure of 9608 pseudocode, it's clear that statements cannot all be used anywhere. For instance, functions and procedures are not first-class, so you cannot declare a function inside a procedure, or vice-versa, or a function in a function, etc. And you obviously cannot use a RETURN statement in a procedure, or anywhere except in a FUNCTION really. And then CASE option statements really only allow a single-line statement, which excludes WHILE/REPEAT/FOR/IF structures ...

So to follow these rules more easily instead of throwing everything to the resolver and interpreter, we are going to need to follow some kind of hierarchy when parsing these statements.


ngjunsiang:
# Statement parsing hierarchy

Let's try to write down these rules as simply as we can:
https://github.com/nyjc-computing/pseudo/blob/122c6445d3ad62fb36ddc6be10b405bf49a00aa2/parser.py#L400-L414

I'll also rearrange the statement parser rules a little to make the next part easier: [e85db895aa616785d41720d368205da5b4a93d39]

ngjunsiang:
And now the part I dread most: coming up with more names for the additional statement-parsing functions we'll need 😩

In the absolute laziest fashion, I'll just keep to `statements` and label the functions `1` to `6`:
https://github.com/nyjc-computing/pseudo/blob/fc6fce650a1e1f04ae7281d68bebb6ae4c74860b/parser.py#L416-L434

And temporarily rename all `statement()` calls to `statement1()`: [fc6fce650a1e1f04ae7281d68bebb6ae4c74860b]

ngjunsiang:
Let's do it:
https://github.com/nyjc-computing/pseudo/blob/4028e66b5e14779f2d51aefb9aa33d124fe23ae2/parser.py#L416-L466

A straightforward implement of the statement-parsing hierarchy described above.

ngjunsiang:
# Applying statement hierarchies

Now we update the statement parsers where appropriate, ensuring they use the correct entry points: [06aa2511684368533d9bb4f80da1a9c51f6a0563]

And remembering to start from entry point 2 in our main parsing loop (since RETURN is not expected in the main program):
https://github.com/nyjc-computing/pseudo/blob/f71b3ba13c028334e831eccd359cc1d294ede7c3/parser.py#L474-L479

ngjunsiang:
And we are done!

PR #45, opened 2022-05-04
18a Boolean

ngjunsiang:
Okay, somehow we have gone 17 chapters and written a working interpreter without incorporating support for booleans. Let's actually do it.

# Scanning booleans

Let's start by adding those tokens to `builtin.py`:
https://github.com/nyjc-computing/pseudo/blob/e8e3cf90b14d358e1dc03638ab3bd0fbfeb1ccc4/builtin.py#L90-L92

While we're it, let's add NULL as well, why not.

And then recognise them in the scanner:
https://github.com/nyjc-computing/pseudo/blob/e8e3cf90b14d358e1dc03638ab3bd0fbfeb1ccc4/scanner.py#L95-L119


ngjunsiang:
We took out the `makeToken() helper not too long ago, but here it can really help us shrinkwrap our code, so let's revamp it:
https://github.com/nyjc-computing/pseudo/blob/39a1ca9a3321fd51968bc90c4ca2713164cc97ed/scanner.py#L21-L24

And use it: [39a1ca9a3321fd51968bc90c4ca2713164cc97ed]

ngjunsiang:
# Parsing booleans

https://github.com/nyjc-computing/pseudo/blob/39a1ca9a3321fd51968bc90c4ca2713164cc97ed/parser.py#L80-L86

Looks like our `literal()` parser can handle booleans without any new code 😎

ngjunsiang:
# Resolving booleans

Our resolver inserts frames and type-checks expressions. None of that involves booleans; Literals already handle that for us.

The work comes in when we work on logical expressions.

ngjunsiang:
# Interpreting booleans

We mostly lean on Python's built-in boolean support, so there's not much to change here as well.

We do have to deal with return values, so we best make sure that we can handle `True` and `False` values correctly:
https://github.com/nyjc-computing/pseudo/blob/3cf37eb1f51c197a6dd67c7aafaadfb6f5fb1cf9/interpreter.py#L51-L52
https://github.com/nyjc-computing/pseudo/blob/3cf37eb1f51c197a6dd67c7aafaadfb6f5fb1cf9/interpreter.py#L79-L80

Since pseudocode uses UPPERCASE, while Python uses Title Case, we do have to be careful that the Output statement casts these values to the appropriate casing:
https://github.com/nyjc-computing/pseudo/blob/3cf37eb1f51c197a6dd67c7aafaadfb6f5fb1cf9/interpreter.py#L82-L87

ngjunsiang:
# Bugfixes

A fix for incorrection statement termination detection: [b3da09ccc1419bfdf0afc883d935f310d3937cab], [8dac349438c0983e527d7a5c32146f74fc96ba5a]

ngjunsiang:
# Error types

Up till now, we are mostly guessing at where the error is (even if we guess pretty well!) It would really help to know what kind of error it is though, so let's include that: [dfb7c97908741aa00d21068bcb458575a45a845b]

ngjunsiang:
And we have boolean support ✔️

PR #48, opened 2022-05-05
18b Logical operators

ngjunsiang:
Now that our interpreter can understand TRUE and FALSE, let's implement AND, OR, and NOT operators.

We start with scanning. First let's add the operators in `builtin.py` and have them recognised as operators:
https://github.com/nyjc-computing/pseudo/blob/4241f231018f605311a67ad762c9db3e4cf5e8cb/builtin.py#L71-L78
https://github.com/nyjc-computing/pseudo/blob/4241f231018f605311a67ad762c9db3e4cf5e8cb/builtin.py#L105-L119

Then enable scanning them:
https://github.com/nyjc-computing/pseudo/blob/4241f231018f605311a67ad762c9db3e4cf5e8cb/scanner.py#L98-L100

### DIFF
--- a/builtin.py
+++ b/builtin.py
@@ -41,7 +41,9 @@ class LogicError(PseudoError):
 def add(x, y):
     return x + y
 
-def sub(x, y):
+def sub(x, y=None):
+    if y is None:
+        return -x
     return x - y
 
 def mul(x, y):
--- a/builtin.py
+++ b/builtin.py
@@ -68,6 +68,15 @@ def ne(x, y):
 def eq(x, y):
     return x == y
 
+def AND(x, y):
+    return x and y
+
+def OR(x, y):
+    return x or y
+
+def NOT(x):
+    return not x
+
 
 
 # Token types
@@ -104,6 +113,9 @@ OPERATORS = {
     '>=': gte,
     '<>': ne,
     '=': eq,
+    'AND': AND,
+    'OR': OR,
+    'NOT': NOT,
 }
 
 SYMBOLS = [',', ':']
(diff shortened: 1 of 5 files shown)


ngjunsiang:
# Operator precedence

NOT has the same precedence as unary `-`, so we parse it at the same line:
https://github.com/nyjc-computing/pseudo/blob/96d886e011e1e6c384f6c56d2546a9d81b7000cd/parser.py#L123-L124

AND and OR have lower precedence than `<>` and `=`; they have the lowest precedence of all the operators. So we add a new parser for them, between equality()` and `expression()`:
https://github.com/nyjc-computing/pseudo/blob/96d886e011e1e6c384f6c56d2546a9d81b7000cd/parser.py#L193-L207

ngjunsiang:
# Resolving logical operators

Next, we've got to resolve these new operators so that appropriate types are returned. First, the unary NOT:
https://github.com/nyjc-computing/pseudo/blob/4deb5a00d98af85be24d26fc5e89ebc028e8549b/resolver.py#L52-L54

Then the binary AND & OR:
https://github.com/nyjc-computing/pseudo/blob/4deb5a00d98af85be24d26fc5e89ebc028e8549b/resolver.py#L59-L62

ngjunsiang:
# Interpreting logical operators

Okay, nothing to change here; our operators take in the )type-checked) input, and return their output. But a slightly horrifying discovery:
https://github.com/nyjc-computing/pseudo/blob/4deb5a00d98af85be24d26fc5e89ebc028e8549b/interpreter.py#L29-L31

For unary `-`, we are passing a single argument to our `sub` operator, which only accepts two arguments 😱:
https://github.com/nyjc-computing/pseudo/blob/4deb5a00d98af85be24d26fc5e89ebc028e8549b/builtin.py#L44-L45

Fix - we overload `sub` to also handle a single argument, by returning its negative:
https://github.com/nyjc-computing/pseudo/blob/763213fc914a2b1cf2a44a2f64a21bc046432343/builtin.py#L44-L47

ngjunsiang:
And we have working logical operators 👍

Issue #7, opened 2022-04-14
Support for logical operators


ngjunsiang:
9608 pseudocode has `AND`, `OR`, and `NOT` logical operators. These take only boolean expressions/values, and always return a boolean value.

Implement support for them in **pseudo**.

Issue #5, opened 2022-04-14
Support for negative values


ngjunsiang:
**pseudo** does not support negative values yet.

Suggestion: implement negative values as an expression with an implicit `0` as left value.

ngjunsiang:
With unary expressions implemented, this is now working.

PR #49, opened 2022-05-05
18c Fix: Logical operators

ngjunsiang:
Oops, one bug that slipped testing! 

Test code:
```
PROCEDURE TestBool(Succeeded : BOOLEAN)
    IF Succeeded = TRUE AND NOT (-1 = -1)
      THEN
        OUTPUT "Yay!"
      ELSE
        OUTPUT "Awww!"
    ENDIF
ENDPROCEDURE

CALL TestBool(TRUE)
```

This doesn't work, because ...
https://github.com/nyjc-computing/pseudo/blob/660d9fd53224954cf78282b4d79d06271ef7f4dc/parser.py#L88-L95

Our `unary()` parser is too generous, parsing an entire expression, which gives us `-(1 = -1)` instead of `((-1) = (-1))`.

Simple fix: parse with `value()` instead of `expression()`:
e4bdb4a9f9beaee3c06baadca6f707016a52bb06

### DIFF
--- a/main.pseudo
+++ b/main.pseudo
@@ -1,5 +1,5 @@
 PROCEDURE TestBool(Succeeded : BOOLEAN)
-    IF Succeeded = TRUE AND NOT 1 = 1
+    IF Succeeded = TRUE AND NOT -1 = -1
       THEN
         OUTPUT "Yay!"
       ELSE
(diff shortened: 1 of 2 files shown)


ngjunsiang:
Another minor issue: errors involving names display a Name Expr instead of a token. That's because we didn't insert the correct token. Fix: [4420dc64dbfed0ab56490cfaf2a7ed8697f691cd]

And with that, we also wrap up two issues: #5 and #7.

PR #52, opened 2022-05-05
20 Packaging

ngjunsiang:
We could go on and implement more features: CHAR, ARRAY, even TYPE!

But first it's time to "graduate" our little project: it's time to make it a Python package!

We'll be using the [poetry package manager](https://python-poetry.org/) to help us with this.

# Project setup

Poetry provides a basic usage page which we'll be using closely: https://python-poetry.org/docs/basic-usage/

Following the page, we run the `poetry new pseudo` command to create our "starter pack", ending up with a new folder and a bunch of files: [0a0cf7df5201c9dbcb05d172bd124bbd72149723]


ngjunsiang:
Let's see what's in this newly created `pseudo` diretory:
- `pseudo` directory, containing `__init__.py`
- `tests` directory
- `pyproject.toml` file
- `README.rst` file

We won't need that double-nesting structure: when someone clones this repository using `git clone`, it is going to put everything in this project into an outer `pseudo` folder, so let's unpack this new directory into our current directory.

Not everything needs to stay though. We have a [README.md](https://github.com/nyjc-computing/pseudo/blob/main/README.md), so we won't need `README.rst`: [785b31478ff59cd2ccb0ffb5a17410409795ae8a]

Next, we edit the `pyproject.toml` file, adding authorship information and dependency information (packages and versions needed by pseudo), and put it in our main folder. I had previously set up a `.gitignore` file so that `git` will ignore other files in my project folder and not push them to the repository; now I have to edit it to allow `pyproject.toml` to be uploaded: [48e16151dc0daa67dc3f4b7a08400c3a34242a5a]

Finally we move the two internal folders, `pseudo` and `tests`, into the main directory: [3e17d592d1611fc936e9d09fd18818dd92565d92]

ngjunsiang:
# Licensing

A part of the project few developers like to care very much about is the licensing. I'm going to pick a permissive one, the MIT License: [1890db25897ee03c9048522038cb30d548da0d59]

ngjunsiang:
# Source files

Since all our Python files (except `main.py`) are part of the `pseudo` package, it is time to move them into the `pseudo` folder: [ec6c2364e32e457954b01214fcf1268f96bc7c89]

The `__init__.py`, even though it is empty, tells Python that its parent `pseudo` directory is a package directory. That means we can do `import pseudo` and `from pseudo import ...`. We shall do this in `main.py`, updating our imports: [f27adbc92c10be206efa54e521421d2808f1b9b9]
https://github.com/nyjc-computing/pseudo/blob/f27adbc92c10be206efa54e521421d2808f1b9b9/main.py#L4-L8

Within `pseudo`, in our source files, the Python files still need to import each other. Previously, they were all in the **main directory** and considered individual single-file packages. Now, they are all in a **package directory**, so we need to use **relative import** syntax (see https://realpython.com/absolute-vs-relative-python-imports/ for more information): [3fa85e659de400804cdba1f6378ce32aeef8a08f]

And we're back to working code again ✨

ngjunsiang:
# Using pseudo

When a user (i.e. pseudocode programmer) uses pseudo, they are going to `import pseudo` and use some kind of entry point to run their code. Better still, installing the `pseudo` package from poetry will add a `pseudo` executable which they can run from the commandline (like `$ pseudo main.pseudo`) and pass the name of their pseudocode source file.

But we're not there yet; for now, let's see what we can do to get there. The first thing we need to do is get all that code out of `main.py`; it should begin to look like what a reasonable user would write, and we certainly wouldn't expect a user to know that much about the scanning-parsing-resolving-interpreting pipeline!

Let's move that code into `pseudocode/__init__.py` instead: [9b765d9c2cac855b303cd2b04d91c4b4d4313df0]

Notice that whatever objects are instantiated in `pseudo/__init__.py`, including functions, become available through the `pseudo` package name when it is imported. The code that used to live in `main()` in `main.py` now lives in `pseudo.run()` (The `run()` function in `pseudo` package).

ngjunsiang:
Let's make it a little nicer to use; it is reasonable that a user would want to call `pseudo.run('mypseudocode.pseudo')` to run pseudocode in a `mypseudocode.pseudo` file, so let's make it happen:
https://github.com/nyjc-computing/pseudo/blob/b155a8ef248241e0f309f7de4ad6492237645b76/pseudo/__init__.py#L16-L18

ngjunsiang:
We could add more features that make `pseudo` nicer to use, but let's leave that to later chapters; this chapter is focused on making a package and publishing it.

ngjunsiang:
# Publishing

The moment of truth! Is our package ready to deploy to [PyPI](https://pypi.org)? It will never be 100% ready, but we have to publish at some point. Let's see if our package name is already taken: https://pypi.org/search/?q=pseudo

Oops, there's another package with the same name. We'll have to publish under a different name:
https://github.com/nyjc-computing/pseudo/blob/2ef419b3067c8a7ef56e769edc2cb0bcd385475d/pyproject.toml#L1-L2

And then we build the package again:

```
~/pseudo$ poetry build
Building pseudo-9608 (0.1.0)

  ModuleOrPackageNotFound

  No file/folder found for package pseudo-9608
```

It didn't work 😰

[... some googling later ...]

Apparently, if we want to publish a package ("pseudo-9608") with a name different from the directory ("pseudo"), we'll have to include this information in `pyproject.toml` under `packages`: https://python-poetry.org/docs/pyproject/#packages

So let's do that:
https://github.com/nyjc-computing/pseudo/blob/b04d0994fd3ac1cbda304cb021336f8c10f06c00/pyproject.toml#L6-L8

```
~/pseudo$ poetry build
Building pseudo-9608 (0.1.0)
  - Building sdist
  - Built pseudo-9608-0.1.0.tar.gz
  - Building wheel
  - Built pseudo_9608-0.1.0-py3-none-any.whl
```

And now it works!

ngjunsiang:
publish this package:

```
~/pseudo-9608$ poetry publish
Username: 
```

Oops, I need to register an account on PyPI. With that done:

```
~/pseudo$ poetry publish
Username: <redacted>
Password: 
Publishing pseudo-9608 (0.1.0) to PyPI
 - Uploading pseudo-9608-0.1.0.tar.gz 100%
 - Uploading pseudo_9608-0.1.0-py3-none-any.whl 100%
```

And it's published! 😎  https://pypi.org/project/pseudo-9608/

ngjunsiang:
Let's wrap up by adding a bit more project information into pyproject.toml`:
https://github.com/nyjc-computing/pseudo-9608/blob/5e7332b08adb862eb07101e2d5e978238a997d2b/pyproject.toml#L1-L12

PR #53, opened 2022-05-06
21a Test: Data passing

ngjunsiang:
Now that we have a package, it's unconscionable to not have tests to be run before building the package.

We will need test cases, which need to be able to verify the output and also any errors raised.

Let's start simple: Separate out the error-handling code.
https://github.com/nyjc-computing/pseudo-9608/blob/ef1a123f3eb42ff5bdb20087992c4a20f3123e48/pseudocode/__init__.py#L15-L23

Then we can use it when errors are raised:
https://github.com/nyjc-computing/pseudo-9608/blob/56aa514e89708575f44aea9b7e069b5cabd29a48/pseudocode/__init__.py#L32-L34
https://github.com/nyjc-computing/pseudo-9608/blob/56aa514e89708575f44aea9b7e069b5cabd29a48/pseudocode/__init__.py#L37-L39

### DIFF
--- a/pseudo
+++ b/pseudo
@@ -13,9 +13,9 @@ Pseudo {pseudocode.__version__}
 
 if __name__ == "__main__":
     srcfile = "main.pseudo"
-    print(HELP)
     if len(sys.argv) > 1:
         srcfile = sys.argv[1]
+    # print(HELP)
     # print("""Usage: pseudo [FILE]""")
     # sys.exit(65)
 
--- a/pseudo
+++ b/pseudo
@@ -19,5 +19,5 @@ if __name__ == "__main__":
     # print("""Usage: pseudo [FILE]""")
     # sys.exit(65)
 
-    pseudocode.run(srcfile)
+    pseudocode.runFile(srcfile)
     
--- a/pseudo
+++ b/pseudo
@@ -18,6 +18,6 @@ if __name__ == "__main__":
         srcfile = sys.argv[1]
     # print("""Usage: pseudo [FILE]""")
     # sys.exit(65)
-    breakpoint()
+
     pseudocode.run(srcfile)
     
--- a/pseudocode/__init__.py
+++ b/pseudocode/__init__.py
@@ -35,5 +35,5 @@ def run(srcfile):
     try:
         frame = interpreter.interpret(statements, frame)
     except RuntimeError as err:
-        print(err)
+        error(lines, err)
         sys.exit(70)
--- a/pseudocode/__init__.py
+++ b/pseudocode/__init__.py
@@ -12,6 +12,16 @@ __version__ = '0.2.1'
 
 
 
+def error(lines, err):
+    errType = type(err).__name__ + ':'
+    if err.line:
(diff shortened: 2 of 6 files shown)


ngjunsiang:
Let's set up a test. We put some pseudocode in the `tests` folder:
https://github.com/nyjc-computing/pseudo-9608/blob/01a954de65d116bd8a7dfef5d1b5915b87628d80/tests/procedure.pseudo

And use the `unittest` module (https://docs.python.org/3/library/unittest.html) to write a test for PROCEDURE:
https://github.com/nyjc-computing/pseudo-9608/blob/01a954de65d116bd8a7dfef5d1b5915b87628d80/tests/test_procedure.py#L5-L13

Hmm ... how are we going to pass the `src` string to our interpreter though?

We could just pass the file directly, but it would likely to helpful in the near future to enable our interpreter to take in strings as well:
https://github.com/nyjc-computing/pseudo-9608/blob/c97194b804a7762bcabc6634f0243cf5a7017050/pseudocode/__init__.py#L25-L32

And now we can pass the source code string to the interpreter:
https://github.com/nyjc-computing/pseudo-9608/blob/236ca069ae3480a263a92c20d3c141d69018f751/tests/test_procedure.py#L7-L16

ngjunsiang:
What next? How would we know if it worked? How would the test case know if an error was raised, or if a value was correctly stored?

We would need to capture data and check it in the test case. We could potentially capture:
- tokens
- statements
- frame
- errors
- output

We probably won't need tokens and statements, since those are internal to the application and the test should not care about the implementation. But the other three would be quite useful to test. Let's capture them in an object; let's start simple and use a dict:
https://github.com/nyjc-computing/pseudo-9608/blob/3ef98f150ceccd8fe2d6761236cddfa3e5c22a29/pseudocode/__init__.py#L30-L35
https://github.com/nyjc-computing/pseudo-9608/blob/3ef98f150ceccd8fe2d6761236cddfa3e5c22a29/pseudocode/__init__.py#L40-L41
https://github.com/nyjc-computing/pseudo-9608/blob/3ef98f150ceccd8fe2d6761236cddfa3e5c22a29/pseudocode/__init__.py#L44-L45

ngjunsiang:
Now we can add test cases to examine the frame data:
https://github.com/nyjc-computing/pseudo-9608/blob/55c1d4f50b3cb7ace53cb8ac310bf650f9be9b1f/tests/test_procedure.py#L13-L30

Test result:
```
$ python -m unittest
Yay!
.
Ran 1 test in 0.001s

OK
```

Great!

That "Yay!" is looking out of place though ... is there a way we can capture output separately?

ngjunsiang:
# Output handlers

https://github.com/nyjc-computing/pseudo-9608/blob/55c1d4f50b3cb7ace53cb8ac310bf650f9be9b1f/pseudocode/interpreter.py#L92-L98

Currently, we execute OUTPUT statements by directly printing them. We'll need to be able to customise this behaviour to capture output data instead of printing it.

PR #54, opened 2022-05-06
21b Test: Checking output

ngjunsiang:
Continuing where we left off, let's upgrade our interpreter to a class so we can register different output handlers.

# Interpreter class

Introducing the Interpreter class:
https://github.com/nyjc-computing/pseudo-9608/blob/b86a88779ac52b1fcdef79c01eaaccd0e2e7d95b/pseudocode/interpreter.py#L36-L46

And using it in `pseudocode`:
https://github.com/nyjc-computing/pseudo-9608/blob/9f7c6b16750766825d39662cddc75c0c35e4a862/pseudocode/__init__.py#L7
https://github.com/nyjc-computing/pseudo-9608/blob/9f7c6b16750766825d39662cddc75c0c35e4a862/pseudocode/__init__.py#L46-L48

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,3 +58,4 @@ This project is also an attempt to write a programming book in a new style. Each
 - [18c Fix: logical operators](https://github.com/nyjc-computing/pseudo/pull/49) (This is an addendum to 18b)
 - [19 REALs](https://github.com/nyjc-computing/pseudo/pull/51)
 - [20 Packaging](https://github.com/nyjc-computing/pseudo-9608/pull/52)
+- [21a Test: Data passing](https://github.com/nyjc-computing/pseudo-9608/pull/53)
--- a/pseudo
+++ b/pseudo
@@ -13,9 +13,9 @@ Pseudo {pseudocode.__version__}
 
 if __name__ == "__main__":
     srcfile = "main.pseudo"
-    print(HELP)
     if len(sys.argv) > 1:
         srcfile = sys.argv[1]
+    # print(HELP)
     # print("""Usage: pseudo [FILE]""")
     # sys.exit(65)
 
--- a/pseudo
+++ b/pseudo
@@ -20,7 +20,8 @@ if __name__ == "__main__":
     # print("""Usage: pseudo [FILE]""")
     # sys.exit(65)
 
-    result = pseudocode.runFile(srcfile)
+    pseudo = pseudocode.Pseudo()
+    result = pseudo.runFile(srcfile)
     lines = result['lines']
     err = result['error']
     if err:
--- a/pseudo
+++ b/pseudo
@@ -2,6 +2,7 @@
 import sys
 
 import pseudocode
+from pseudocode.builtin import ParseError, LogicError, RuntimeError
 
 
 
@@ -19,5 +20,13 @@ if __name__ == "__main__":
     # print("""Usage: pseudo [FILE]""")
     # sys.exit(65)
 
-    pseudocode.runFile(srcfile)
-    
+    result = pseudocode.runFile(srcfile)
+    lines = result['lines']
+    err = result['error']
+    if err:
+        if type(err) in (ParseError, LogicError):
+            pseudocode.error(lines, err)
+            sys.exit(65)
+        elif type(err) in (RuntimeError,):
+            pseudocode.error(lines, err)
+            sys.exit(70)
--- a/pseudocode/__init__.py
+++ b/pseudocode/__init__.py
@@ -4,7 +4,7 @@ from pseudocode.builtin import ParseError, RuntimeError, LogicError
 from pseudocode import scanner
 from pseudocode import parser
 from pseudocode import resolver
-from pseudocode import interpreter
+from pseudocode.interpreter import Interpreter
 
 
 
@@ -43,8 +43,9 @@ def run(src):
         sys.exit(65)
     else:
         result['frame'] = frame
+    interpreter = Interpreter(frame, statements)
     try:
-        frame = interpreter.interpret(statements, frame)
+        frame = interpreter.interpret()
     except RuntimeError as err:
         result['error'] = err
         error(lines, err)
--- a/pseudocode/__init__.py
+++ b/pseudocode/__init__.py
@@ -12,44 +12,55 @@ __version__ = '0.2.1'
 
 
 
-def error(lines, err):
-    errType = type(err).__name__ + ':'
-    if err.line:
-        lineinfo = f"[Line {err.line}]"
-        print(lineinfo, lines[err.line - 1])
-    if err.col:
-        leftmargin = len(lineinfo) + 1 + err.col
-        print((' ' * leftmargin) + '^')
-    print(errType, err.report())
+class Pseudo:
+    """A 9608 pseudocode interpreter"""
+    def __init__(self):
+        self.handlers = {
+            'output': None,
+        }
 
-def runFile(srcfile):
-    with open(srcfile, 'r') as f:
-        src = f.read()
-    run(src)
+    def registerHandlers(self, output):
+        self.handlers['output'] = output
+
+    @staticmethod
+    def error(lines, err):
+        errType = type(err).__name__ + ':'
+        if err.line:
+            lineinfo = f"[Line {err.line}]"
+            print(lineinfo, lines[err.line - 1])
+        if err.col:
+            leftmargin = len(lineinfo) + 1 + err.col
+            print((' ' * leftmargin) + '^')
+            print(errType, err.report())
+
+    def runFile(self, srcfile):
+        with open(srcfile, 'r') as f:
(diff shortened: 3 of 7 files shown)


ngjunsiang:
# Registering output handler

Now we can add a `registerOutputHandler()` method, with a default handler that uses Python’s `print()`:
https://github.com/nyjc-computing/pseudo-9608/blob/7b003630dee0bb220c2de13345b7656a3b7d962e/pseudocode/interpreter.py#L36-L40

https://github.com/nyjc-computing/pseudo-9608/blob/7b003630dee0bb220c2de13345b7656a3b7d962e/pseudocode/interpreter.py#L45-L51

ngjunsiang:
# Using output handler

Now we'll need some way to use it in `execOutput()`. With OOP, we generally have two options:
- make `execOutput()` an Interpreter method so it can access object attributes
- pass additional arguments to `execOutput()` to customise its output-handling behaviour

ngjunsiang:
## Overloading execOutput()

https://github.com/nyjc-computing/pseudo-9608/blob/2bc55ef5b0ab612d837a35624f15bd15d249722b/pseudocode/interpreter.py#L117-L126

ngjunsiang:
https://github.com/nyjc-computing/pseudo-9608/blob/f4df601b8eda669bedb9b175fa79f14d4d29a8b1/pseudocode/interpreter.py#L53-L59

https://github.com/nyjc-computing/pseudo-9608/blob/f4df601b8eda669bedb9b175fa79f14d4d29a8b1/pseudocode/interpreter.py#L115-L119

https://github.com/nyjc-computing/pseudo-9608/blob/f4df601b8eda669bedb9b175fa79f14d4d29a8b1/pseudocode/interpreter.py#L216-L218

https://github.com/nyjc-computing/pseudo-9608/blob/f4df601b8eda669bedb9b175fa79f14d4d29a8b1/pseudocode/lang.py#L271-L275

ngjunsiang:
When running code for a user, the interpreter is perfectly fine using `print()` and `input()` as output and input handlers. But when testing the interpreter, we'll need a different output handler, one that captures and stores the output for checking later.

We'll also need a different function to retrieve that stored data. Why not just pass an object around? Because that will make it difficult later to change how storage and retrieval works, and I don't know what additional features may take us to change our design decisions.

For now I imagine it'll look something like this:

https://github.com/nyjc-computing/pseudo-9608/blob/1edbc0c9329dcfbee92c057db819c17fb985821d/tests/__init__.py#L7-L24

Ideally, `captureOutput()` will be useable interchangeably with `print()`, and `returnData()` will require no arguments. How will they access the `storage` object (here, we use a dict) then? It'll need to be kept around, but not in global space, because that leaves it open to mutation (i.e. editing) by other functions.

We'll put it in a closure, i.e. a function's scope that is kept around after exit. So we wrap `storage`, `captureOutput()`, and `returnData()` in this function—let's call it `capture()`:

https://github.com/nyjc-computing/pseudo-9608/blob/1edbc0c9329dcfbee92c057db819c17fb985821d/tests/__init__.py#L5-L28

When `capture()` is called, it returns a capture-return function pair, which can be used to capture data and retrieve it respectively.

I anticipate we might need other kinds of capture-return functions, so we use a `captureType` parameter to decide what to return.

ngjunsiang:
Let's implement `captureOutput()`:

https://github.com/nyjc-computing/pseudo-9608/blob/e3c3d9c4812e80ea9d3335419544f49ce8b4f97f/tests/__init__.py#L16-L27

We remove the arguments that `captureOutput()` can't handle (and which we don't use) anyway. And store the strings instead of `print()`ing them.

ngjunsiang:
# Capturing output in testing

Now let's use `capture()` in testing. I imagine it'll look something like this:

https://github.com/nyjc-computing/pseudo-9608/blob/fa429efeddc5a5b1e78199d69ddb93543b638c36/tests/test_procedure.py#L11-L18

Oh dear. We enabled the interpreter to register an output handler ... but we use `pseudocode.run()` in testing, without direct access to the interpreter. Looks like we'll need to make a Pseudo class to wrap our scan-parse-resolve-interpret pipeline too?

The entire process would look like this:

https://github.com/nyjc-computing/pseudo-9608/blob/fa429efeddc5a5b1e78199d69ddb93543b638c36/tests/test_procedure.py#L9-L18

(No, `pseudocode.Pseudo` doesn't actually exist yet! Doesn't stop us from trying to imagine how it might be used; it'll help us with the design.)

We call `capture()` to get our capture-return function pair. Register the capturing function as the output handler for our Pseudo object. Call the retrieval function later to retrieve the captured data and store in `result`.

ngjunsiang:
# Testing output

Finally we can imagine how to check the captured outdoor in a different test:

https://github.com/nyjc-computing/pseudo-9608/blob/beea3eff47d296edd3de58a5e9b842e6831934c1/tests/test_procedure.py#L39-L41

ngjunsiang:
# The Pseudo class

I haven't forgotten! We'll wrap `run()` and `runFile()` in this class, and give it a handler-registering method:

https://github.com/nyjc-computing/pseudo-9608/blob/6b3c9bc7b9e75e3a33bb300a9de4b66d4d4ddbf4/pseudocode/__init__.py#L15-L66

We'll also throw in `error()` since that needs to be used; this will probably be refactored later but it's a good start. It's defined as a static method since it doesn't require any access to the Pseudo object (and we wouldn't want that too).

ngjunsiang:
# Default handlers

Let's initialise Pseudo with its default handlers:

https://github.com/nyjc-computing/pseudo-9608/blob/89e5512020a6eba090e48963f466f092e074c995/pseudocode/__init__.py#L27-L33

This is how we can access our default error handler without needing another Pseudo method:

https://github.com/nyjc-computing/pseudo-9608/blob/89e5512020a6eba090e48963f466f092e074c995/pseudocode/__init__.py#L43-L44
https://github.com/nyjc-computing/pseudo-9608/blob/89e5512020a6eba090e48963f466f092e074c995/pseudocode/__init__.py#L54-L57

Just get it from `self.handlers`! (This also overrides the global `error()` function so we don't accidentally end up using it and being confused.)

ngjunsiang:
# Registering handlers

Since we have a `handlers` attribute that stores handlers, `registerHandlers()` can check in there for existing keys; we won't need to define explicit keyword arguments, and can rely on the more generic `kwargs` parameter:

https://github.com/nyjc-computing/pseudo-9608/blob/1559bc96e38a21ad5cb2973651cdee00566467e4/pseudocode/__init__.py#L36-L40

ngjunsiang:
And let's remove error-handling code from Pseudo. We'll edit it to focus on returning the result of interpreting code, whether there's an error or not:

https://github.com/nyjc-computing/pseudo-9608/blob/7e309c18e7bd4d286cf18cc78cceb706715dbe2c/pseudocode/__init__.py#L46-L68

ngjunsiang:
Back in `pseudo`, remember to use the new Pseudo class:
https://github.com/nyjc-computing/pseudo-9608/blob/53051d8ded9b6f4fc4f0bcfd43929899c510650c/pseudo#L23-L24

And return the result of `run()` in `runFile()`:
https://github.com/nyjc-computing/pseudo-9608/blob/53051d8ded9b6f4fc4f0bcfd43929899c510650c/pseudocode/__init__.py#L41-L44

ngjunsiang:
# Bugfixing

Test code:
```
PROCEDURE TestBool(Succeeded : BOOLEAN)
    IF Succeeded = TRUE AND NOT (-1.2 = -1.5)
      THEN
        OUTPUT "Yay!"
      ELSE
        OUTPUT "Awww!"
    ENDIF
ENDPROCEDURE

CALL TestBool(TRUE)
```

Oops, we forgot that to get to OUTPUT, we have to pass through ProcFunc and If Stmts as well. We should just pass `kwargs` through every Stmt at this point. And then through the Calls as well, which means passing `kwargs` through `evalCall`, which means passing them through the `Expr.accept()` method ... that's a whole bunch of changes: [e2348dd61bc7495f122c9324d194b455f0d4d623], [9241f2a7f4128df5157aeaad4af7ad4c435bc3b5], [a37d29bebc4840fdc8cdbcca060a5fb99589e48f], [a37d29bebc4840fdc8cdbcca060a5fb99589e48f]

Oops, our test failed ... a brain bug. Let's fix that test condition: [1c816ad7b35477c322f66bf192cc1fe501c24d98]

# Cleanup

Since default handlers are being set in Pseudo, we no longer need Output to come up with its own default. Let's remove that: [e49dd59985f68f60b7c547fcd68e8bbb42f4b8a9]

And then a quick cleanup: [0b8e76a3fd4f63bfcea89d3dbb8ff1191c69d93b], [be6344e0aa539e0d83269cc10226c471c9795f75]

ngjunsiang:
And now we can test program output in automated fashion!

PR #55, opened 2022-05-07
21c Test: Checking Errors

ngjunsiang:
Besides checking that our code successfully executes what should be executed, we also want to check that our code raises errors where it should. These are captured in the result object (which we'll have to turn into a class soon, before the result interface gets locked into a huge body of test code).


ngjunsiang:
# Checking error-free code

We start by checking our existing test code did not raise any errors:
https://github.com/nyjc-computing/pseudo-9608/blob/cc2200942cfe8f926d19406420730c8f07fc47cf/tests/test_procedure.py#L30-L31

# Test design

To keep our tests simple to maintain, let's bundle the test code together with its script instead of putting it in a separate file:
https://github.com/nyjc-computing/pseudo-9608/blob/cc2200942cfe8f926d19406420730c8f07fc47cf/tests/test_procedure.py#L6-L17
https://github.com/nyjc-computing/pseudo-9608/blob/cc2200942cfe8f926d19406420730c8f07fc47cf/tests/test_procedure.py#L26

Now we have one less file, and skip file-reading issues. This way we can also design our tests so that each test script tests one code block.

ngjunsiang:
# Checking error-raising code

Let's add a `test_error.py` test:
https://github.com/nyjc-computing/pseudo-9608/blob/1dd3fb6ec45c18b09dfbaaa452c092f5e83f374a/tests/test_error.py#L5-L12

This should definitely raise a `ParseError`, if our code is working fine!

So let's get the error from `result` and check that it's an error of the correct type:
https://github.com/nyjc-computing/pseudo-9608/blob/1dd3fb6ec45c18b09dfbaaa452c092f5e83f374a/tests/test_error.py#L14-L26

And that's it! We'll be adding more test cases like these when we encounter code that should return a particular kind of error, but a bug in our code causes a different kind of error to be raised.

PR #56, opened 2022-05-08
22a Scoping: Recursion

ngjunsiang:
The next thing I want to do actually is to implement some built-in functions, like `RND()` and `RANDOMBETWEEN()`. But something more pressing awaits: while pseudo alloss us to declare FUNCTIONs and PROCEDUREs in global space, we can't actually call them from within a PROCEDURE or FUNCTION yet.

And that is because so far, the local scopes only declare whatever parameters they are defined with. We have yet to introduce the wider notion of global scope, where any names declared are available to all function/procedure calls. And we'll need this because ... where else are we going to put built-in functions?

And, in an oblique roundabout way, we are going to approach this by tackling recursion first. After all, recursion involves a procedure/function calling itself, which has to first be declared in the global scope right? Solve that and almost everything else will work.

### DIFF
--- a/pseudocode/lang.py
+++ b/pseudocode/lang.py
@@ -126,7 +126,8 @@ class Frame:
     delete(name)
         deletes the slot associated with the name
     """
-    def __init__(self):
+    def __init__(self, outer=None):
+        self.outer = outer
         self.data = {}
 
     def __repr__(self):
--- a/pseudocode/lang.py
+++ b/pseudocode/lang.py
@@ -61,6 +61,30 @@ class Procedure(Callable):
 
 
 
+class Builtin(Value):
+    """
+    Represents a system function in pseudo.
+
+    Attributes
+    ----------
+    - params
+        A list of parameters used by the callable
+    - func
+        the Python function to call when invoked
+    """
+    __slots__ = ('params', 'func')
+    def __init__(self, params, func):
+        self.params = params
+        self.func = func
+
+    def __repr__(self):
+        attrstr = ", ".join([
+            repr(getattr(self, attr)) for attr in self.__slots__
+        ])
+        return f'{type(self).__name__}({attrstr})'
+
+
+
 class File(Value):
     """
     Represents a file object in a frame.
@@ -312,15 +336,6 @@ class Input(Stmt):
     def __init__(self, rule, name):
         self.rule = rule
         self.name = name
-
-
-
-# class Assign(Stmt):
-#     __slots__ = ('rule', 'name', 'expr')
-#     def __init__(self, rule, name, expr):
-#         self.rule = rule
-#         self.name = name
-#         self.expr = expr
 
 
 
--- a/pseudocode/lang.py
+++ b/pseudocode/lang.py
@@ -125,6 +125,8 @@ class Frame:
         updates the value associated with the name
     delete(name)
         deletes the slot associated with the name
+    lookup(name)
+        returns the first frame containing the name
     """
     def __init__(self, outer=None):
         self.outer = outer
@@ -157,6 +159,12 @@ class Frame:
 
     def delete(self, name):
         del self.data[name]
+
+    def lookup(self, name):
+        if self.has(name):
+            return self
+        if self.outer:
+            return self.outer.lookup(name)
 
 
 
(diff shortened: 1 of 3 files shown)


ngjunsiang:
Since we have a testing framework using `unittest` now, let's use that to add a test for recursion:

Test code:
https://github.com/nyjc-computing/pseudo-9608/blob/c9076f8d7f13f1b30b9b1a0796ece76b8ee63334/tests/test_recursion.py#L7-L15

The tests:
https://github.com/nyjc-computing/pseudo-9608/blob/c9076f8d7f13f1b30b9b1a0796ece76b8ee63334/tests/test_recursion.py#L28-L38

ngjunsiang:
# Accessing globals

Now, one way to resolve this without adding new features is to iterate through `global` names, declaring and assigning them in `local` if they don't exist there. This is fragile and very prone to breaking (what happens if the globals are updated?)

Instead, let's extend `Frame` so that it is aware of an `outer` frame:
https://github.com/nyjc-computing/pseudo-9608/blob/54c439928f3c4a419b70ee57c2048e2cac0fa070/pseudocode/lang.py#L129-L131

What can we do with an `outer` though? For starters, our resolver needs to do frame insertion. Each verifier function is given a frame to work with, but the Get exprs it has to resolve might not find the name they were looking for in that given frame. When that happens, we can check in `outer` to see if the name is there. And in the case of recursion, or nested calls, we can even follow the entire chain of `outer`s until we find the frame with that name declared.

Let's add a `lookup()` method to do that:
https://github.com/nyjc-computing/pseudo-9608/blob/db9d81987fa41face8aafa53f15eb3dc4f81f105/pseudocode/lang.py#L163-L167

ngjunsiang:
# Resolving in outer

Now let's make this happen in the resolver: we want to insert the frame containing the name when we resolve a Get expr. That happens in the `resolveGet()` resolver:
https://github.com/nyjc-computing/pseudo-9608/blob/959864aa9ab7ac0c79993652aaa55761ae183cb7/pseudocode/resolver.py#L111-L119

And we use our shiny new `lookup(0` method to do that, if the initial `frame` does not have `expr.name` declared.

# Recursive calls

Next, `resolveProcCall()` and `resolveFuncCall()`:
https://github.com/nyjc-computing/pseudo-9608/blob/cdc6a97e9381bee80d88c8c49bfbd3143caeae9c/pseudocode/resolver.py#L121-L137

After we correctly insert the frame (with `expr.callable.accept(frame, resolveGet)`), we have to retrieve this frame before we `getValue()` our procedure/function (because the `frame` argument there is the `local` frame).

This still doesn't work yet, because in `verifyFunction()` and `verifyProcedure()` we verify the nested statements before we've properly declared and assigned the function/procedure:
https://github.com/nyjc-computing/pseudo-9608/blob/cdc6a97e9381bee80d88c8c49bfbd3143caeae9c/pseudocode/resolver.py#L231-L252

We'll have to rearrange things a little to make it work:
https://github.com/nyjc-computing/pseudo-9608/blob/05946eeac2fad4c893bac780309967ad44990c56/pseudocode/resolver.py#L208-L229
https://github.com/nyjc-computing/pseudo-9608/blob/05946eeac2fad4c893bac780309967ad44990c56/pseudocode/resolver.py#L231-L252

The workflow here is:
1. Setup the `local` frame first, with the original `frame` as its `outer`.
2. Declare the callable name in the outer `frame`
3. Initialise the Procedure/Function with `local`, and assign it to the name in `frame`.

ngjunsiang:
# Resolving recursive calls

A few more minor tweaks and fixes before this will work:
https://github.com/nyjc-computing/pseudo-9608/blob/bfb5957335855b2c906d61259f5086138ec76ee0/pseudocode/resolver.py#L139-L147

Because the resolver isn't concerned with retrieving values (that is the interpreter's work), and we don't have any helper functions/functions/evaluators to do that, we'll have to do it manually:
1. Get the frame from `Call.callable`
2. Retrieve the `callable` object from the frame (using `Call.callable.name`)

## Test fixes

And then a couple of fixes for our test code: a wrong procedure call, then some jankiness around terminal line breaks (which we'll `strip()` away since they are not the key focus for the test): [bfb5957335855b2c906d61259f5086138ec76ee0]

And now our recursion tests pass!

PR #57, opened 2022-05-08
22b Scoping: System

ngjunsiang:
Let's work our way towards the `RND()` and `RANDOMBETWEEN()` functions, which we shall here call **system functions**, because they cannot be implemented using pseudocode. What are they going to look like?

# Comparing Callables and system functions

As Python functions, probably much like the following:
https://github.com/nyjc-computing/pseudo-9608/blob/bf9a173472f3ba0bf1762c03a367be9786c81e7f/pseudocode/system.py#L1-L11

These look nothing like pseudo Callables though:
https://github.com/nyjc-computing/pseudo-9608/blob/bf9a173472f3ba0bf1762c03a367be9786c81e7f/pseudocode/lang.py#L26-L44

Let's compare them:
- While a Callable does its work in a (`local`) frame, system functions do their work in another language, usually the language that the interpreter is implemented in. Here, that's Python.
- Both Callables and system functions have parameters.
- Both Callables and system functions have code to execute; a list of statements for the former, and a Python function for the latter.


ngjunsiang:
# System functions

Let's define a new class for system functions; we'll call them `Builtin`s.

https://github.com/nyjc-computing/pseudo-9608/blob/bf9a173472f3ba0bf1762c03a367be9786c81e7f/pseudocode/lang.py#L64-L84

ngjunsiang:
We'll also have to treat them differently in `evalCall()`:
https://github.com/nyjc-computing/pseudo-9608/blob/40e87ec5a6de43606d685d08626de35abfa6cf71/pseudocode/interpreter.py#L82-L95

Builtins, being essentially a wrapper around a Python function, are much easier to evaluate: just call the function. This feels a bit clunky, we'll see if we can use polymorphism to make it more elegant later on.

ngjunsiang:
# The system frame

It's no use just having Builtins and being able to call them in the interpreter ... we need to be able to retrieve them from a frame. But which frame?

We could declare and assign them in the global `frame` ... but that feels hacky and risky. Users could accidentally override them, or declare a new variable over them, and we would have to figure out ways to prevent/mitigate that ... or we could just make `system` the `outer` frame of the global frame. I like that.

First let's use relative imports instead of absolute imports:
[1e9a1d2b7139bb6d37a06b8ad7b4e477216b30e5]

Now we can pull in the `system` frame, and initialise a `globalFrame` with that as the outer:
https://github.com/nyjc-computing/pseudo-9608/blob/10099aca477688224388b6b7a9467dfabf03010d/pseudocode/__init__.py#L51
But ... where will we pass it to the resolver?

ngjunsiang:
# Using the system frame in Resolver

I imagine that once it's implemented (though we haven't written a line of code for that yet), it'll look something like this: [6c87fe74000f420d2b74b1b21b3bf79440b6c6d3]

Looks kinda messy, let's clean it up a little. There's so much code in the `try-except` blocks, let's separate it out into recognisable stages:
https://github.com/nyjc-computing/pseudo-9608/blob/41e082d7a08eaa27763cd46af10ad8421b56ee2a/pseudocode/__init__.py#L46-L78

`globalFrame` only needs to be assigned into `result['error']` once since it is still the same object after passing through the pipeline; we might as well do it at the top.

We return the result whenever an error occurs, otherwise we store the result.

And now we just have to create the Resolver class.

ngjunsiang:
# The Resolver

Here it is, after some minor debugging:
https://github.com/nyjc-computing/pseudo-9608/blob/4fbb375292d54491d470aa487e2ae5bd9195c9ac/pseudocode/resolver.py#L50-L59

And now we can remove the `resolve()` function: [73884aadfa4eeef25054aefd2cf1770b41498053]

ngjunsiang:
# Testing

Time to test the code. Let's run some pseudocode in `main.pseudo` first:

```
OUTPUT RND()
OUTPUT RANDOMBETWEEN(2, 9)
```

There was some error ... which didn't display. Because of a typo that caused no error messages to appear, if there's no `col` info. Let's fix that: [a9fe9f0398fb8589cef17d3e99211df590e8ad46]

Ah, our parser can't recognise argument-less calls (such as `RND()`). Let's fix that in the parser, then enable our resolver to recognise Builtins: [f5ab7d95d75c205748c59b165cc1587ece2d044a]

ngjunsiang:
And now our tests pass 😎

ngjunsiang:
While we're working on system functions, let's add one more shall we?

# Adding EOF

`EOF(filename)` returns TRUE if `filename` has been opened (with OPENFILE) and its current position is not at the end of the file yet.

Let's take it out of our keyword list, so it can get recognised as a name: [544aba369b035b770d431a4e082d1cc345ddd063]

Then we add the function itself:
https://github.com/nyjc-computing/pseudo-9608/blob/544aba369b035b770d431a4e082d1cc345ddd063/pseudocode/system.py#L33-L42

And declare it in `system`:
https://github.com/nyjc-computing/pseudo-9608/blob/544aba369b035b770d431a4e082d1cc345ddd063/pseudocode/system.py#L43-L49

ngjunsiang:
# Interpreting EOF

The sole argument to EOF is a string, representing the name of an open file. But the EOF system function expects a Python file handle. So we'll need some code that evaluates the string, and uses it to retrieve the file handle from the frame.

In our interpreter, within the code for executing Builtins, we add a check for EOF:
https://github.com/nyjc-computing/pseudo-9608/blob/7469a31fada5d45dad7b46eefecaf0a4ae6a4d4c/pseudocode/interpreter.py#L84-L90

ngjunsiang:
# Testing EOF

Lastly, let's add a test case. We'll set up some pseudocode for testing:
https://github.com/nyjc-computing/pseudo-9608/blob/68fb2c4cc778e49db6f8a297b84ce6a7402465f5/tests/test_system.py#L8-L14

This requires a test file, `"testfile.txt"` to exist. Let's make it happen in `setUp()`:
https://github.com/nyjc-computing/pseudo-9608/blob/68fb2c4cc778e49db6f8a297b84ce6a7402465f5/tests/test_system.py#L17-L21

The rest of the setup instantiates Pseudo and interprets the code, capturing its output like we did before:
https://github.com/nyjc-computing/pseudo-9608/blob/68fb2c4cc778e49db6f8a297b84ce6a7402465f5/tests/test_system.py#L22-L28

Then we check the output:
https://github.com/nyjc-computing/pseudo-9608/blob/68fb2c4cc778e49db6f8a297b84ce6a7402465f5/tests/test_system.py#L34-L39

It's not nice to leave unnecessary files lying around after testing. We can use `TestCase.tearDown()` to carry out any post-test cleanup:
https://github.com/nyjc-computing/pseudo-9608/blob/68fb2c4cc778e49db6f8a297b84ce6a7402465f5/tests/test_system.py#L41-L42

ngjunsiang:
And with the tests passed, we are done 👌

PR #58, opened 2022-05-11
23a Object: Scopes

ngjunsiang:
I debated a while whether to add this as chapter 22c, since it involves scopes, but eventually decided that objects should get their own chapter.

Now that we have implemented scoping, allowing procedure/function statements to access global space, let's look at situations where we might not want a frame to be able to do those things. Let's talk about Objects.

# What is an object?

In most languages, objects represent a bundle of data. This data might be stored in the object with integer indexes, and we usually call these objects lists/sequences. The data might also be stored with string keys/attributes, and we call these mappings or structs or objects.

Whatever we call them, these usually serve as data containers, without holding any code (at least not until we get to object-oriented programming). That means there won't be any code that needs access to the `outer` scope, and means Objects won't need to implement `outer` lookups.

Secondly, at the most basic level we do not expect to delete indexes and attributes in a statically typed language like 9608 pseudocode.

We could still use a Frame for this, and just be careful not to do those things ... but this feels like a different kind of container, even if it will still need many of Frame's methods. The use of frames involves lots of validation, while that's not the case for Objects.

# The Object class

So let's define an Object class, and then have Frame extend it through inheritance:
https://github.com/nyjc-computing/pseudo-9608/blob/dfb5a27e03196868d7d400211bd6102ea4c3184b/pseudocode/lang.py#L128-L177

https://github.com/nyjc-computing/pseudo-9608/blob/dfb5a27e03196868d7d400211bd6102ea4c3184b/pseudocode/lang.py#L181-L204


ngjunsiang:
# Defining custom types

In 9608 pseudocode, custom types are defined with the following code:

```
TYPE <identifier1>
    DECLARE <identifier2> : <data type>
    DECLARE <identifier3> : <data type>
    ...
ENDTYPE
```

So let's get these keywords recognised:
https://github.com/nyjc-computing/pseudo-9608/blob/41524b4db1edac5bee32835593f8f6b2d83f7816/pseudocode/builtin.py#L86-L96

And let's create a new type of Stmt for them:
https://github.com/nyjc-computing/pseudo-9608/blob/41524b4db1edac5bee32835593f8f6b2d83f7816/pseudocode/lang.py#L387-L390

Hmm ... we only expect DECLARE statements in a TYPE block, and don't want to accidentally parse other kinds of statements. We could hardcode that, or else extend our statement hierarchy (#44), which I am loathe to do, so let's just hardcode it. That means we'll have a list of Declare exprs instead of Stmts:
https://github.com/nyjc-computing/pseudo-9608/blob/2ac58e2c16fd2f7195dd94a36adf495a6b897af2/pseudocode/lang.py#L387-L390

ngjunsiang:
# Scanning custom type statements

We added the keywords to the scanner, and no further changes are needed 😌

ngjunsiang:
# Parsing custom type statements

A TYPE statement is essentially a declaration; we can use it wherever we use DECLARE, so let's put them in the same statement hierarchy:
https://github.com/nyjc-computing/pseudo-9608/blob/267170bebaa20ebbd6165803cee908328fdfc7d4/pseudocode/parser.py#L458-L463

And add a parser for them:
https://github.com/nyjc-computing/pseudo-9608/blob/267170bebaa20ebbd6165803cee908328fdfc7d4/pseudocode/parser.py#L245-L246

Oops, we forgot about the `rule` argument for TypeStmt:
https://github.com/nyjc-computing/pseudo-9608/blob/267170bebaa20ebbd6165803cee908328fdfc7d4/pseudocode/lang.py#L387-L391

ngjunsiang:
Now let's test the parsing:

```
TYPE Mytype
    DECLARE Aaa : INTEGER
    DECLARE Bbb : INTEGER
ENDTYPE
```

And implement it too:
https://github.com/nyjc-computing/pseudo-9608/blob/ce47033974af19703d6c6e40589d9bc2953dbc4b/pseudocode/parser.py#L245-L255

ngjunsiang:
Let's see some example code from the pseudocode guide for testing:

```
TYPE Student
    DECLARE Surname : STRING
    DECLARE FirstName : STRING
    DECLARE DateOfBirth : DATE
    DECLARE YearGroup : INTEGER
    DECLARE FormGroup : CHAR
ENDTYPE
DECLARE Pupil1 : Student
Pupil1.Surname <- "Johnson"
Pupil1.Firstname <- "Leroy"
Pupil1.DateOfBirth <- 02/01/2005
Pupil1.YearGroup <- 6
Pupil1.FormGroup <- ꞌAꞌ
```

After we declare a type, we have to be able to use it by declaring other variables as *that type*. How are we going to achieve this?

Right now, our scanner will recognise `Student` in `DECLARE Pupil1 : Student` as a `name` token. That means our parser will parse it as a Name Expr:
https://github.com/nyjc-computing/pseudo-9608/blob/ce47033974af19703d6c6e40589d9bc2953dbc4b/pseudocode/parser.py#L121-L137

ngjunsiang:
# Registering types

Currently our typechecks essentially involve checking a set of type keywords (`'INTEGER'`, `'STRING'`, etc) to see if a particular word is found in the set. This set is hardcoded in `builtin`:
https://github.com/nyjc-computing/pseudo-9608/blob/5beb760448337ec2f51df8f3910fc5568dff1517/pseudocode/builtin.py#L104-L108

We need:
1. A way to add to this list of types as new types are registered, for type-checking purposes
2. A way to retrieve the type information, i.e. the declared parameters in custom TYPEs
3. A way to transfer type information from object to object, e.g. when the resolver hands over to the interpreter

This sounds kind of similar to Frame; could we extend Frame to include these responsibilities?
https://github.com/nyjc-computing/pseudo-9608/blob/5beb760448337ec2f51df8f3910fc5568dff1517/pseudocode/lang.py#L180-L184

This is where docstrings come in handy for documenting design intent. Here I write pretty clearly that Frame manages TypedValues. Could we represent types as TypedValues? We could use `'TYPE'` as each type's type, and ... you know what, this is just really messy. I think this could be a good idea in other dynamically typed languages, where each type is represented as a factory/constructor function, but I don't think it's a good fit for 9608 pseudocode.

Let's make (yet) another class instead.

ngjunsiang:
# The TypeSystem

We start very simple, essentially a wrapper around a dictionary:
https://github.com/nyjc-computing/pseudo-9608/blob/3838b3eeaa273c84e20af631e1331bbd3d8a66c2/pseudocode/lang.py#L112-L143

We allow initial registration of types (since we will start off with a set of built-in types). Declaring a new type just adds another name to the internal dict. For complex types, we allow an optional template that will be used to create new objects of that type. (An interesting edge case, which we won't consider now, is what happens if a user tries to declare nested custom types?)

ngjunsiang:
# Initialising TypeSystem

Now that we have TypeSystem, where should we initialise it?

We could initialise it in the same place as frame, and pass it around together with the frame ... which would involve editing a lot of function calls, with very little payoff. What if we just embed it in the frame for now?
https://github.com/nyjc-computing/pseudo-9608/blob/c799e7eb504719864578218b3effc86c05e1d5ad/pseudocode/lang.py#L215-L231

Looks okay ... but this won't transfer to `local` frames, and possibly implies that each frame could have its own set of types. It is unclear if scoped types are supported in 9608 pseudocode, but they don't seem to be necessary, and in any case I'll put my foot down here and decide not to explicitly support them. Let's have `types` inherited from `outer` when inited, if possible:
https://github.com/nyjc-computing/pseudo-9608/blob/1c205a79461f8672c54995768861f7e7762e1eee/pseudocode/lang.py#L228-L234

ngjunsiang:
# Using TypeSystem

In our Resolver class, let's get the ball rolling:
https://github.com/nyjc-computing/pseudo-9608/blob/f9938d3ab15cc51fc72da55fe8f97fd1df9c09c0/pseudocode/__init__.py#L47-L48

In one line (and a couple of imports) we have rolled our typesystem into the syntax tree.

Now let's pick up where we left off: after parsing type statements, we resolve them.

ngjunsiang:
# Resolving TYPE declarations

Inside a TYPE-ENDTYPE block, programmers can only DECLARE attributes that the custom type has. This makes things a bit easier for us; we only parsed DECLAREs, instead of having to choose a `statementN()` entry point. So we are reasonably confident that if the parser did its job bug-free, we will only have Declare Exprs to handle:

https://github.com/nyjc-computing/pseudo-9608/blob/e872954c25eac22d13d3badc94a589fa1ac56db6/pseudocode/resolver.py#L280-L285

Because we were careful to make Object and Frame polymorphic (implementing the same methods with the same interface), we can re-use `resolveDeclare()` with our Object here, to declare the attributes inside the Object.

And then we declare a new type in the frame. Object will not be used to directly store TypedValues. Instead, whenever a user declares a new variable with this type, we will use this Object as a template to make another one.

ngjunsiang:
This isn't adequate for custom types, since we lose all information about what attributes it should have. Instead of instantiating a TypedValue, we should make a copy of the template Object in `Frame.types`:
https://github.com/nyjc-computing/pseudo-9608/blob/21b851fc034dad3c5b5c16eb64eaabc69193814b/pseudocode/lang.py#L198-L200

ngjunsiang:
Not forgetting to pass `typesys` to the object template when we instantiate it in `verifyDeclareType()`:
https://github.com/nyjc-computing/pseudo-9608/blob/325b8dfce2fcabc22cfc45ea8f14a072e3e3dbef/pseudocode/resolver.py#L282

ngjunsiang:
# Copying templates

e use

https://github.com/nyjc-computing/pseudo-9608/blob/e229f47f59e7a69b1a21c3a104d6d2353839278e/pseudocode/lang.py#L224-L230

ngjunsiang:
We implicitly expect TypeSystem to have a ready template for each type stored in it; we can no longer instantiate the type names with a `None` value. Let's update TypeSystem's init:
https://github.com/nyjc-computing/pseudo-9608/blob/dda5d03341e39936b100ce11fff5f19cfff70627/pseudocode/lang.py#L130-L134

ngjunsiang:
Feels clunky and kinda more repetitive than it needs to be, but it's a start.

ngjunsiang:
# Cleaning up

Some things that need to be fixed before we can move on: Frames can't be initialised without a system, which means `system` needs a typesystem, so we might as well stuff our built-in types there instead of initialising them in Pseudo:
https://github.com/nyjc-computing/pseudo-9608/blob/2bf5007990842bbb86931f5c69781df812ec120b/pseudocode/system.py#L8

Changes: [2bf5007990842bbb86931f5c69781df812ec120b], [19011e648e973d44287c19948e851879435d3be9]

ngjunsiang:
# Testing

And finally we get to the point where we see this:

`ParseError: [Line 7 column 6] <None> '.': Expected <- after name`

Now we're getting somewhere; it's still an error, but it's an error where our parser hasn't figured out what `.` is supposed to be and what to do with it. And that's because we haven't gotten to Gets for object attributes ... which will have to be a different chapter.

PR #59, opened 2022-05-18
23b Object: Attributes

ngjunsiang:
Before starting this chapter, a very minor delayed update in our parser for parsing precedence: [dccbb97e97a32ab08684eba773b361343689ce02]

----------

After we register our custom type and enable variables to be declared with this type, how do we ... you know, retrieve its attributes or update them?

Most languages use a dot(`.`) notation for accessing object attributes, and 9608 pseudocode is no different. `Student.Name` refers to the `Name` attribute of `Student`. Let's start by getting our scanner to recognise them.

# Scanning attributes

Oh, looks like our scanner already recognises the `.` symbol and can tokenise them just fine. Moving on to the parser.

### DIFF
--- a/main.pseudo
+++ b/main.pseudo
@@ -4,9 +4,12 @@ TYPE Student
     DECLARE YearGroup : INTEGER
 ENDTYPE
 DECLARE Pupil1 : Student
+DECLARE Var : INTEGER
 Pupil1.Surname <- "Johnson"
 Pupil1.FirstName <- "Leroy"
 Pupil1.YearGroup <- 6
+Var <- 9
 OUTPUT Pupil1.Surname
 OUTPUT Pupil1.FirstName
 OUTPUT Pupil1.YearGroup
+OUTPUT Var
--- a/main.pseudo
+++ b/main.pseudo
@@ -5,5 +5,5 @@ TYPE Student
 ENDTYPE
 DECLARE Pupil1 : Student
 Pupil1.Surname <- "Johnson"
-Pupil1.Firstname <- "Leroy"
+Pupil1.FirstName <- "Leroy"
 Pupil1.YearGroup <- 6
--- a/main.pseudo
+++ b/main.pseudo
@@ -7,3 +7,6 @@ DECLARE Pupil1 : Student
 Pupil1.Surname <- "Johnson"
 Pupil1.FirstName <- "Leroy"
 Pupil1.YearGroup <- 6
+OUTPUT Pupil1.Surname
+OUTPUT Pupil1.FirstName
+OUTPUT Pupil1.YearGroup
--- a/pseudocode/interpreter.py
+++ b/pseudocode/interpreter.py
@@ -105,8 +105,7 @@ def evalCall(frame, expr, **kwargs):
 
 def evalAssign(frame, expr):
     value = expr.expr.accept(frame, evaluate)
-    obj = expr.assignee
-    # evaluate assignee until object is retrieved
+    obj = evaluate(frame, expr.assignee)
     obj.setValue(expr.name, value)
 
 def evaluate(frame, expr):
--- a/pseudocode/interpreter.py
+++ b/pseudocode/interpreter.py
@@ -77,7 +77,9 @@ def evalBinary(frame, expr):
 def evalGet(frame, expr):
     # Frame should have been inserted in resolver
     # So ignore the frame that is passed here
-    return expr.frame.getValue(expr.name)
+    obj = expr.frame
+    # evaluate obj until object is retrieved
+    return obj.getValue(expr.name)
 
 def evalCall(frame, expr, **kwargs):
     callable = expr.callable.accept(frame, evalGet)
@@ -100,7 +102,9 @@ def evalCall(frame, expr, **kwargs):
 
 def evalAssign(frame, expr):
     value = expr.expr.accept(frame, evaluate)
-    frame.setValue(expr.name, value)
+    obj = expr.assignee
+    # evaluate assignee until object is retrieved
+    obj.setValue(expr.name, value)
 
 def evaluate(frame, expr):
     if isinstance(expr, Literal):
--- a/pseudocode/interpreter.py
+++ b/pseudocode/interpreter.py
@@ -1,5 +1,5 @@
 from .builtin import RuntimeError
-from .lang import Frame, File, Callable, Builtin
+from .lang import Object, File, Callable, Builtin
 from .lang import Literal, Unary, Binary, Get, Call, Assign
 from .system import EOF
 
@@ -78,11 +78,12 @@ def evalGet(frame, expr):
     # Frame should have been inserted in resolver
     # So ignore the frame that is passed here
     obj = expr.frame
-    if isinstance(obj, Object):
-        return obj.getValue(expr.name)
     # evaluate obj until object is retrieved
     if type(obj) in (Get, Call):
-        return evaluate(frame, obj)
+        obj = evaluate(frame, obj)
+    if not isinstance(obj, Object):
+        raise RuntimeError("Invalid object", expr.frame.token())
+    return obj.getValue(expr.name)
 
 def evalCall(frame, expr, **kwargs):
     callable = expr.callable.accept(frame, evalGet)
@@ -105,10 +106,10 @@ def evalCall(frame, expr, **kwargs):
 
 def evalAssign(frame, expr):
     value = expr.expr.accept(frame, evaluate)
-    obj = evaluate(frame, expr.assignee)
+    obj = evaluate(frame, expr.assignee.frame)
     obj.setValue(expr.name, value)
 
-def evaluate(frame, expr):
+def evaluate(frame, expr, **kwargs):
     if isinstance(expr, Literal):
         return expr.accept(frame, evalLiteral)
     if isinstance(expr, Unary):
--- a/pseudocode/interpreter.py
+++ b/pseudocode/interpreter.py
@@ -78,8 +78,11 @@ def evalGet(frame, expr):
     # Frame should have been inserted in resolver
     # So ignore the frame that is passed here
     obj = expr.frame
+    if isinstance(obj, Object):
+        return obj.getValue(expr.name)
     # evaluate obj until object is retrieved
-    return obj.getValue(expr.name)
+    if type(obj) in (Get, Call):
+        return evaluate(frame, obj)
 
 def evalCall(frame, expr, **kwargs):
     callable = expr.callable.accept(frame, evalGet)
--- a/pseudocode/interpreter.py
+++ b/pseudocode/interpreter.py
@@ -106,7 +106,9 @@ def evalCall(frame, expr, **kwargs):
 
 def evalAssign(frame, expr):
     value = expr.expr.accept(frame, evaluate)
-    obj = evaluate(frame, expr.assignee.frame)
+    obj = expr.assignee.frame
+    if type(obj) in (Get, Call):
+        obj = evaluate(frame, obj)
     obj.setValue(expr.name, value)
 
 def evaluate(frame, expr, **kwargs):
--- a/pseudocode/lang.py
+++ b/pseudocode/lang.py
@@ -131,7 +131,14 @@ class TypeSystem:
         self.data = {}
         for typeName in types:
             self.declare(typeName)
-            self.setTemplate(typeName, TypedValue(typeName, None))
+            self.setTemplate(typeName, None)
+
+    def __repr__(self):
+        nameTypePairs = [
+            f"{name}: {self.getTemplate(name)}"
+            for name in self.data
+        ]
+        return f"{{{', '.join(nameTypePairs)}}}"
 
     def has(self, name):
         return name in self.data
@@ -140,7 +147,7 @@ class TypeSystem:
         self.data[name] = None
 
     def setTemplate(self, name, template):
-        self.data[name] = template
+        self.data[name] = TypedValue(name, template)
 
     def getTemplate(self, name):
         return self.data[name]
@@ -209,10 +216,17 @@ class Object:
         return name in self.data
 
     def declare(self, name, type):
-        template = self.types.getTemplate(type)
-        self.data[name] = template.copy()
+        if name == 'Pupil1':
+            breakpoint()
+        try:
+            template = self.types.getTemplate(type)
+            self.data[name] = template.copy()
+        except:
+            breakpoint()
 
     def getType(self, name):
+        if type(self.data[name]) is not TypedValue:
+            breakpoint()
         return self.data[name].type
 
     def getValue(self, name):
@@ -229,7 +243,7 @@ class Object:
         Class = type(self)
         newobj = Class(typesys=self.types)
         for name in self.data:
-            newobj.declare(name, newobj.getType(name))
+            newobj.declare(name, self.getType(name))
         return newobj
 
 
--- a/pseudocode/lang.py
+++ b/pseudocode/lang.py
@@ -326,10 +326,11 @@ class Declare(Expr):
 
 
 class Assign(Expr):
-    __slots__ = ('name', 'expr')
-    def __init__(self, name, expr, token=None):
+    __slots__ = ('name', 'assignee', 'expr')
+    def __init__(self, name, assignee, expr, token=None):
         super().__init__(token=token)
         self.name = name
+        self.assignee = assignee
         self.expr = expr
 
 
--- a/pseudocode/lang.py
+++ b/pseudocode/lang.py
@@ -249,6 +249,8 @@ class Frame(Object):
 
     Methods
     -------
+    set(name, typedValue)
+        assigns the given TypedValue to the name
     delete(name)
         deletes the slot associated with the name
     lookup(name)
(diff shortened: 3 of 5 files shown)


ngjunsiang:
# Getting an attribute

Okay, we can do this in a number of ways. Our object is very similar to a Frame, except without `outer`-chaining and deletion of names. And we already have an Expr for retrieving named values from a Frame: that's our Get Expr. We could just reuse Get, maybe ...?

Or we could make a new Expr, let's call it Attr. We would need a new resolver and evaluator for it.

Hmm, without trying to plan further than I can imagine, I think I'll go with reusing a Get Expr. Resolving is gonna be a bit tricky, since we can't just insert the frame; we'll have to retrieve the object and insert that instead, and the resolver can't do that (since the resolver doesn't carry out any instantiation of objects). It will have to be the interpreter's job.

Let's clean up some things before we start parsing attribute gets:
- updating documentation in `value()`: [e0e08c29a6088d14cc52a1b8d48844ee0c7e6055]
- initialise data structures within `if`: [a78a52151557c4f583acf688d5b6fa5804ac29df]

ngjunsiang:
# Parsing attribute gets

Let's start with thinking through how to proceed:
https://github.com/nyjc-computing/pseudo-9608/blob/bacf9a40092d32f81829dcddfc302888c23f0f65/pseudocode/parser.py#L120-L130

`nameExpr()` already starts with a Get Expr that retrieves the typedvalue with the name (presumably a custom type object, in the case of attribute gets). If we encounter a `.`, we make another Get Expr:
https://github.com/nyjc-computing/pseudo-9608/blob/c845787167d514c847da2e9796386ad7da30b20d/pseudocode/parser.py#L120-L129

ngjunsiang:
This code doesn't handle nested attributes yet (e.g. `Student.Name.First`, but that's a tiny fix for us: [7b7da657f9610e8891933021414eacf95b7fe2dd]

ngjunsiang:
Now comes some major refactoring: up to this point, we've been assuming that Assign Exprs only ever need to store a name and an expr. We assume that this name is declared in the frame. But if we are assigning values to an attribute, that's a name declared in an object and not in the frame.

We'll have to refactor Assign to accept left-hand Gets.

# Assigning to attributes

ngjunsiang:
# Assigning to attributes

We'll need to replace `name` in Assign Exprs with a more general term; let's call it an `assignee`. We'll introduce it without removing `name` yet, the goal being to keep the code working:
https://github.com/nyjc-computing/pseudo-9608/blob/f0eae5ce02c6691ccde46096b8df3d6eb0c7b425/pseudocode/lang.py#L328-L334

And update `makeExpr()` as well to accept an `assignee` argument:
https://github.com/nyjc-computing/pseudo-9608/blob/f0eae5ce02c6691ccde46096b8df3d6eb0c7b425/pseudocode/parser.py#L25-L39

ngjunsiang:
Now we turn our attention to `assignment()`, the first place that creates Assign Exprs:
https://github.com/nyjc-computing/pseudo-9608/blob/f2766dbeeb922273b29e3a6c3f22af89be045ad8/pseudocode/parser.py#L228-L232

Let's make it create and pass an assignee:
https://github.com/nyjc-computing/pseudo-9608/blob/0bff81563e6b8d6115a44acb667033bf4c08026c/pseudocode/parser.py#L228-L241

ngjunsiang:
Finally, our `name()`, `callExpr()` and `attrExpr()`:
https://github.com/nyjc-computing/pseudo-9608/blob/0cb05d8030b04c4b3e3e373cd223b6bd15933ef3/pseudocode/parser.py#L104-L131

ngjunsiang:
And back to `assignment()`: Instead of just parsing name assignments, let's handle attribute assignments as well:
https://github.com/nyjc-computing/pseudo-9608/blob/070c6c4d11c3fa2e9bb296a9bc6a2cd8d4f342d8/pseudocode/parser.py#L233-L245

No more `ParseError`; the next error occurs in the resolver, and we'll get to that shortly.

Note that `assignment()` won't be able to handle expressions like `Name(...).Attribute`; I don't foresee a need for this yet, but if we run across pseudocode like that we can always come back to this and extend it.

ngjunsiang:
We can't remove `name` from Assign yet, not while the resolver and interpreter depend on it. So let's get to resolving next.

# Resolving attribute assignments

Two things we need to handle:
1. Get Exprs potentially having a Get Expr as a `frame`
2. Assign Exprs having an additional `assignee` attribute

First, nested Gets:
https://github.com/nyjc-computing/pseudo-9608/blob/ce8a61c1273cea37068e8f710241aef2e3a700a6/pseudocode/resolver.py#L127-L141
We do a check for a Get Expr, and have the nested `frame` `accept()` `resolveGet()` recursively. When we finally hit the innermost Get Expr (which would be the first one parsed by `name()`), we carry out a frame insertion by checking for `frame is NULL`.

And we lean on the frame's typesystem to help us determine the value's type.

And now the assignee:
https://github.com/nyjc-computing/pseudo-9608/blob/ce8a61c1273cea37068e8f710241aef2e3a700a6/pseudocode/resolver.py#L119-L125
The assignee is going to be a Get Expr, so let's resolve it; thankfully `resolveGet()` now handles nested Gets (although we haven't tested this yet). Then we can do a type-check for the assignee's type and the expr's type.

ngjunsiang:
Let's give it a spin:
```
  File "lang.py", line 228, in getType
    return self.data[name].type
AttributeError: 'Object' object has no attribute 'type'
```

Oops, looks like somehow we stored an Object in the Frame data instead of a TypedValue.

Examining the frame's typesystem in the Python debugger, `pdb`:
```
(Pdb) self.types
{BOOLEAN: <BOOLEAN: None>, INTEGER: <INTEGER: None>, REAL: <REAL: None>, STRING: <STRING: None>, FILE: <FILE: None>, Student: {Surname: STRING, FirstName: STRING, YearGroup: INTEGER}}
```

Everything else except Student seems to be a TypedValue. I'd point the blame at `verifyDeclareType()`, where the Student type is being registered in the type system:
https://github.com/nyjc-computing/pseudo-9608/blob/ce8a61c1273cea37068e8f710241aef2e3a700a6/pseudocode/resolver.py#L287-L292

`verifyDeclareType()` delegates everything to the TypeSystem though, so back there we go:
https://github.com/nyjc-computing/pseudo-9608/blob/ce8a61c1273cea37068e8f710241aef2e3a700a6/pseudocode/lang.py#L142-L143

There's our problem: `setTemplate()` does not do any wrapping of the template. So let's change that:
https://github.com/nyjc-computing/pseudo-9608/blob/1bd2ba8c19c89c6462cfe924164d63a39aaa5900/pseudocode/lang.py#L149-L150

And then we no longer need to manually wrap the types in `__init__()`:
https://github.com/nyjc-computing/pseudo-9608/blob/1bd2ba8c19c89c6462cfe924164d63a39aaa5900/pseudocode/lang.py#L130-L134

ngjunsiang:
Another error:
```
LogicError: <function name at 0x7f3432a741f0>: Expected Student, is STRING
```

First let's fix that `<function name at 0x7f3432a741f0>` which is supposed to be a token: [7a90848dfd78030d81e5b50834a06c7ef0130fba]

And now let's look into why we have an expected Student type when we should be dealing with builtin types. I suspect this is happening in `resolveAssign()`, so let's put a breakpoint there:
```
> /home/runner/pseudo-9608/pseudocode/resolver.py(124)resolveAssign()
-> expectTypeElseError(
(Pdb) expr.assignee
Get(Get({Pupil1: Student}, 'Pupil1'), 'Surname')
(Pdb) expr.assignee.accept(frame, resolveGet)
'Student'
```

Hmm looking strange ... that nested get should be returning the Surname type which is STRING, not Student. Back to `resolveGet()`:
https://github.com/nyjc-computing/pseudo-9608/blob/1bd2ba8c19c89c6462cfe924164d63a39aaa5900/pseudocode/resolver.py#L132-L134

There's our problem. The inner Get returns type Student, naturally, and the outer Get naively passes that along. So let's pass this inner type through our type system first:
https://github.com/nyjc-computing/pseudo-9608/blob/d5b2e3f3ecf3418673dd3335a0a21432c0b71013/pseudocode/resolver.py#L132-L142

Along the way, we have to remember to check for the existence of each object type or attribute name since that will not be done by the type system's methods.

ngjunsiang:
`LogicError: [Line 8 column 7] <None> 'Firstname': Undeclared attribute`

Good, now it's catching pseudocode typos (9608 pseudocode is supposed to be case-insensitive, but we haven't implemented that yet).

And now we can move on to the interpreter.

ngjunsiang:
First, `evalGet()`:
https://github.com/nyjc-computing/pseudo-9608/blob/ec8d0e7c909489ce6bcfba7a170f71042c532943/pseudocode/interpreter.py#L77-L86

`frame` could be an Expr or a Frame. If it's an Expr, we evaluate it recursively, uncovering the nested Gets layer by layer until we get a Frame. Then we can extract our desire value from this Frame.

Then `evalAssign()`:
https://github.com/nyjc-computing/pseudo-9608/blob/ec8d0e7c909489ce6bcfba7a170f71042c532943/pseudocode/interpreter.py#L107-L110

Minor changes here; instead of assuming that we have a frame, we know we need to do some evaluation. The `assignee` is a Get Expr for the object's attribute, so we need to evaluate that Expr's `frame` to get an Object. Then we use that Object's `setValue()` to carry out the attribute assignment.

And then we can test it:
```
TYPE Student
    DECLARE Surname : STRING
    DECLARE FirstName : STRING
    DECLARE YearGroup : INTEGER
ENDTYPE
DECLARE Pupil1 : Student
Pupil1.Surname <- "Johnson"
Pupil1.FirstName <- "Leroy"
Pupil1.YearGroup <- 6
OUTPUT Pupil1.Surname
OUTPUT Pupil1.FirstName
OUTPUT Pupil1.YearGroup
```

It works!

ngjunsiang:
Oops, missed out one thing:
https://github.com/nyjc-computing/pseudo-9608/blob/d877b6ae8e2c928e673b1c471b850df19f52bee0/pseudocode/interpreter.py#L107-L112

Sometimes the assignee's `frame` is already a valid Frame (e.g. when assigning to a variable instead of attribute), so we don't always want to evaluate it. So we do a check first.

And now we can handle both variable and attribute assignment expressions:
```
TYPE Student
    DECLARE Surname : STRING
    DECLARE FirstName : STRING
    DECLARE YearGroup : INTEGER
ENDTYPE
DECLARE Pupil1 : Student
DECLARE Var : INTEGER
Pupil1.Surname <- "Johnson"
Pupil1.FirstName <- "Leroy"
Pupil1.YearGroup <- 6
Var <- 9
OUTPUT Pupil1.Surname
OUTPUT Pupil1.FirstName
OUTPUT Pupil1.YearGroup
OUTPUT Var
```

Result:
```
Johnson
Leroy
6
9
```

ngjunsiang:
# Cleanup

And then some cleaning up before the next chapter.

Refactoring to use helper functions: [d88d20448954bbc9ce6e9c76544bb759b4102ac1]

ngjunsiang:
I'm not quite happy with how I handled verifyFunction and verifyProcedure in the resolver; it looks unwieldly and too detailed, and it is difficult to see the common parts between them.

In particular I don't like the return type checking being in verifyFunction; it looks like it could easily be handled in verifyStmts even if procedures won't use it. Let's upgrade verifyStmts to do that: [b9eb1d674a2a43486a96d4e9c0fbd62c6faf1493]

and then use it in verifyFunction: [db6ec08d6633b577437bd2b9aa4c765c60f6d462]

together with this change, we also need another way to check for the existence of a Return statement. I do this by just checking for a statement with a `'return'` rule.

ngjunsiang:
Found a bug which didn't come up in testing: verifying BYREF parameters was not being done properly. I was setting the value in the `local` frame's typedvalue slot instead of assigning the slot from the global frame, so I fixed that: [3ee36366c8db042f719dd3189d3d21d3277a13b5]

Frame does not have a `set()` method to make this work, so I made one: [968a055a694488aec68b2942b6c9d7871c195f8c]

ngjunsiang:
It's still not quite how I would like it. Within the resolver, it doesn't make sense to use the `Expr.accept()` method when we can reference the resolver we want with e.g. `resolveDeclare(frame, expr)`. So I reformatted that for easier reading (and marginally faster execution): [d9651aea8a74f4f8d282ae0aeecb383988542235]

Oops, accidentally took out some code I needed. Undoing it: [13e91aadff99e06341c1d4eac38b0f00cfce925c]

Declaring parameters shares some code with declaring variables, but can't use `resolveDeclare()`. But I think it should; we are doing similar things to a frame, after all, though BYREF declarations need to be handled differently. No reason why that can't be done in `resolveDeclare()`; we'll need access to the `outer` frame, which we now have.

So let's upgrade `resolveDeclare()`, and then use it to simplify verifyFunction and verifyProcedure: [63be4f3efbf1e2dcac354250538fb0d6890678d6]

ngjunsiang:
Parameter checking and replacement is the same across both verify* functions. Let's make a helper for that: [23c46672952e94a6dcd0113978643e518cf7d1ec]

ngjunsiang:
We are going to be adding a new object type in the next chapter. That means more `match()` and `check()` use; I don't like how I have to use `check(tokens).word` if I don't want the token to be `consume()`d; I want to use `match()` for all matching operations! I upgraded `match()` to make consuming optional: [8ec5c0e67e97594a7811f04ee3663e7e21511192]

Then refactor the parser to use it in place of `check()`: [54a1746f12ccb33752bed9a8c71e30d07ce15266]

ngjunsiang:
I also use `check(tokens).type == ...` in a few places, which does the same thing as `match()` except using the `type` attribute instead of `word`. Let's make a `matchType()` helper: [3ee7cae06ca5493d7a2b2ba14c4812c8e89017af]

Since this doesn't need to consume tokens, I set `advance=False` by default.

ngjunsiang:
Did a more thorough pass for loops in the parser to catch EOFs: [4f1a240179a24bfa05245fb76d0c29caafc80be8]

Next, some really long lines stand out in the parser:
- `while not atEnd(tokens) and match(...)` occurs 13 times
- `while not atEnd(tokens) and match(...)` occurs 7 times
- `match(...)` occurs 51 times.

This needs some reorganising ...

Let's do this:
- `match*()` helpers will consume and return the token: [3df2562cf8417e0fe6b0ad92f38769b677609dc2]
  - `matchElseError()`
  - `matchWord()` (previously `match()`)
- `expect*()` helpers will not consume the token, and return a boolean: [e9ad41af83f851f9fb4d08455b1265f3ba848310]
  - `expectWord()`
  - `expectType()`

ngjunsiang:
About `atEnd(tokens)` ... we want to check for this in a loop because `match*()` and `expect*()` will not consume an EOF token (unless we actively match it), so we might lock ourselves into a loop.

This feels like it might be a constant source of bugs; it relies on human consistency in using this check whenever there's a loop. I just had to add a whole bunch of them earlier. Since absolute blistering performance is not the goal here, I'd rather have safety. I'll throw in an `atEnd()` check to throw an error if we encounter an unexpected EOF:
https://github.com/nyjc-computing/pseudo-9608/blob/d06205dff3798fd57d0e897139752d452682d9e2/pseudocode/parser.py#L18-L20

And use it in the `match*()` and `expect*()` helpers (except the `ElseError` ones): [3df0ae10b23be437af85ee7b13e826221b6111c0]

Refactoring `matchElseError()` to use `matchWord()`: [a0198aa88305d64ab4c0d5d46e033a3d57cff471]

ngjunsiang:
Finally we can take out the `atEnd()` check in our loops: [26158d39b598e711d20c54cfe9e459bec124cf9a]

Not only do we get more readable loops, we also get more safety knowing that our match and expect helpers help us to check and raise errors. This will need testing too, to ensure it's not raising `Unexpected EOF` unexpectedly.

ngjunsiang:
Almsot got myself into infinite recursion there: [9260d507e55e88f69334706e00ac5e129fc279aa]

Can't have `expectType()` and `atEnd()` calling each other!

ngjunsiang:
It's very clean to have match and expect helpers return a boolean, but it's quite limiting. In many places we `expect()` tokens, and then consume them after that. We'll be able to compress some code if our match and expect helpers return the token for a match, and return `None` otherwise. These are still truthy and falsy respectively; they are treated as `True` and `False` respectively when evaluated as booleans. [5538bed31550fad97731a19e2d79fe4c9127839d]

While compressing code, I accidentally also added a whole bunch of other formatting changes. Including adding a `msg` keyword argument to `matchElseError()`: [d7a7f15dc1ef2a00aa5a880c3ec137fb495fef8e]

`matchElseError()` doesn't follow the new naming convention; it matches on words, so I renamed it to `matchWordElseError()`. Wordier, but more accurate: [9fc5f1bef8c99a5c349a622eba4d78507e7d2b2c]

ngjunsiang:
And then bugfixes.

- wrong return value: [c9a0d04a03bfd23a322bc0752f0649ce79aa9571]
- We no longer need to skip line breaks since that is done before the statement: [baa91b3f4183bfd25ed88ca6329401a1c6e91148] (idk, this might need fixing again in future. We'll see.)
- improved return statement detection: [d63aecdc8027b6e3e12b2854ed42ab49f1ff5e55]

ngjunsiang:
At the end of the day, we have cleaner code, and still mostly working. I think we can move on to arrays.

PR #60, opened 2022-05-20
23c Object: ARRAY

ngjunsiang:
Time to move on to arrays.

But first, some refactoring. Let's get rid of hardcoded symbols in the scanner:
https://github.com/nyjc-computing/pseudo-9608/blob/e49634b9dbc38acad839b5a01cb5e34d85a2e24d/pseudocode/builtin.py#L128-L132
https://github.com/nyjc-computing/pseudo-9608/blob/e49634b9dbc38acad839b5a01cb5e34d85a2e24d/pseudocode/scanner.py#L118-L121
https://github.com/nyjc-computing/pseudo-9608/blob/e49634b9dbc38acad839b5a01cb5e34d85a2e24d/pseudocode/scanner.py#L55-L61

And now we add ARRAY as a token:
https://github.com/nyjc-computing/pseudo-9608/blob/e49634b9dbc38acad839b5a01cb5e34d85a2e24d/pseudocode/builtin.py#L108


ngjunsiang:
Test code: `https://github.com/nyjc-computing/pseudo-9608/blob/9b57cd09a44444fb2e02308a1108a280c5b2a251/main.pseudo#L1`

Result: `ParseError: [Line 1 column 23] <None> '[': Expected ('\n',) after statement`

Obviously this doesn't work; our DECLARE expects single TYPE words only.

ngjunsiang:
# Extending Declare

For now we dump additional information under a new attribute, `metadata`, to describe additional info about the type:
https://github.com/nyjc-computing/pseudo-9608/blob/2b3a151b7c3d1569a261031834376bfb531c1d52/pseudocode/lang.py#L331-L337

I think I'll use a dict for the metadata, looking something like:
```
{
   size: [3, 6],
    type: 'INTEGER',
}
```
This describes a 2D array with 3x6 subelements, each of type INTEGER.

ngjunsiang:
Let's make this work for non-arrays first. First we get `makeExpr()` to pass the metadata along: [f113159d7e9b68ad398c7f5d82b44d8f5f1d4b48]

Then have `declare()` pass empty metadata for non-ARRAY types: [79bb50fa614f114b73bf4d06184a935e5df0a145]

Let's add another helper: [c6a402e19802b2e2fbdacd87f9bf0393acab2699]

Then we parse this metadata for arrays:
https://github.com/nyjc-computing/pseudo-9608/blob/b60f37f0803dd78169c983661ed199e63febfc02/pseudocode/parser.py#L296-L319

Just ... lots of matching. If we detect an ARRAY typetoken, we go into ARRAY-parsing mode. Between `[]`, we check for a `start:end` range. We can make the resolver's job easier by checking for `INTEGER`-type tokens using our new helper, `matchTypeElseError()`, for `start` and `end`.

This code can be made more readable with a `colonRange()` helper:
https://github.com/nyjc-computing/pseudo-9608/blob/ec92bda157ada4419f414c8b507882d35e09888e/pseudocode/parser.py#L289-L294

With the changes looking like this: [ec92bda157ada4419f414c8b507882d35e09888e]

ngjunsiang:
I don't like the repeated typetoken check in `declare()`. I'll use an internal helper instead: [466d0034459a56ef8a0ee074b6268b41220c9fa2]

ngjunsiang:
# Resolving ARRAY declarations

Our Declare Expr for the array looks like `Declare('AnArray', 'ARRAY', {'size': [(1, 10)], 'type': 'INTEGER'})`. We can use this information to set up the variable in the frame in our resolver.

ngjunsiang:
## Instantiating the array

`resolveDeclare()` already declares our ARRAY in the frame. Beyond that, our array needs a space to hold its elements; we'll use an object for that. So we also need to set up the slots in the object for those elements. Let's bring the object into existence first:
https://github.com/nyjc-computing/pseudo-9608/blob/ff07efdd8e59da40592263b09c0e8475ccc94587/pseudocode/resolver.py#L75-L79

The tricky thing about 9608 pseudocode arrays: they can be 1D, 2D, or more. We could do some checking to decide how many layers of list nesting to use ... I think I'll skip that. Instead I'll use a little-known trick: tuples as dict keys. We are of course declaring our array as an Object so it shouldn't be relying on dict features like that ... but we can always change that later. At least tuples are a convenient way to represent N-dimensional indexes.

https://github.com/nyjc-computing/pseudo-9608/blob/38c62dae89d47a7a3fc6e4b760935c4afd4d6995/pseudocode/resolver.py#L78-L90

The metadata contains `(start, end)` tuples so we start by converting them to proper ranges. Then use [itertools.product](https://docs.python.org/3/library/itertools.html) to iterate through all possible combinations of the ranges, returning them as tuples.

We declare each index tuple in the array Object (as a TypedValue), with the Declare's metadata type.

ngjunsiang:
Our next problem: parsing array gets (i.e. `AnArray[1]`)

We check for `[` after a name:
https://github.com/nyjc-computing/pseudo-9608/blob/0da60ecd921c68f85df12b743febea27bc5d1d15/pseudocode/parser.py#L188-L197

And then parse comma-delimited numbers until we encounter a `]`:
https://github.com/nyjc-computing/pseudo-9608/blob/0da60ecd921c68f85df12b743febea27bc5d1d15/pseudocode/parser.py#L161-L170

ngjunsiang:
# Parsing array assignments

Very similar to array gets; we add another rule for `[`:
https://github.com/nyjc-computing/pseudo-9608/blob/bb07ca25fe3af51cadf01abadaaad0196841bab9/pseudocode/parser.py#L275-L291

# Testing arrays in the resolver

And then it's testing time:
```
DECLARE AnArray : ARRAY[1:10] OF INTEGER
TYPE Student
    DECLARE Surname : STRING
    DECLARE FirstName : STRING
    DECLARE YearGroup : INTEGER
ENDTYPE
DECLARE Pupil1 : Student
DECLARE Var : INTEGER

FOR Var <- 1 TO 10
    AnArray[5] <- Var
ENDFOR
Pupil1.Surname <- "Johnson"
Pupil1.FirstName <- "Leroy"
Pupil1.YearGroup <- 6
Var <- 9
OUTPUT Pupil1.Surname
OUTPUT Pupil1.FirstName
OUTPUT Pupil1.YearGroup
OUTPUT Var
```

# Bugfixes

Already we run into some typos: [bb7c72b4ee47e3a21a7d25241825cb553c62ea61]

ngjunsiang:
Another problem: the resolver tries to look up the index in the type system, since that is what it does with objects. And there is nothing in the type system on array indexes. This needs a massive fix, full details here: [5a7ecf28c3180c7b6f186fa0c5da7e4a2edfb965]

The broad overview:
https://github.com/nyjc-computing/pseudo-9608/blob/5a7ecf28c3180c7b6f186fa0c5da7e4a2edfb965/pseudocode/resolver.py#L150-L159

If we reach a Get from the global frame (which is still NULL), we insert the frame:
https://github.com/nyjc-computing/pseudo-9608/blob/5a7ecf28c3180c7b6f186fa0c5da7e4a2edfb965/pseudocode/resolver.py#L160-L165

If the frame is a Get, we resolve it recursively:
https://github.com/nyjc-computing/pseudo-9608/blob/5a7ecf28c3180c7b6f186fa0c5da7e4a2edfb965/pseudocode/resolver.py#L166-L169

We check the return type of the evaluated Get. If it's not one of our built-in types, we presume it's a custom type. We then check (a) the existence of this type, (b) the existence of the requested attribute:
https://github.com/nyjc-computing/pseudo-9608/blob/5a7ecf28c3180c7b6f186fa0c5da7e4a2edfb965/pseudocode/resolver.py#L170-L182

Arrays, on the other hand, are not checked in the typesystem but in the frame (which should have been inserted earlier):
https://github.com/nyjc-computing/pseudo-9608/blob/5a7ecf28c3180c7b6f186fa0c5da7e4a2edfb965/pseudocode/resolver.py#L183-L187

Other built-in non-array types are ignored for now; their declared type is returned in the last line:
https://github.com/nyjc-computing/pseudo-9608/blob/5a7ecf28c3180c7b6f186fa0c5da7e4a2edfb965/pseudocode/resolver.py#L188-L190

ngjunsiang:
# Assigning to variable indexes

Let's try this code (snippet shown): [c6baefe30438c20b2bb98e768ec012a1de6de23f]
```
FOR Var <- 1 TO 10
    AnArray[Var] <- Var
ENDFOR
```

Result: `ParseError: [Line 11 column 12] <None> 'Var': Expected ('INTEGER',)`

Mmm, our parsing code only parses integer literals, and can't understand variables. Guess we'll need to make it more flexible: [426a8e416744bf86635e66c82e2e51f7e07d85d3]
https://github.com/nyjc-computing/pseudo-9608/blob/426a8e416744bf86635e66c82e2e51f7e07d85d3/pseudocode/parser.py#L161-L170

Since our indexes might now be an Expr, we won't know their value until runtime; that is the interpreter's job. We'll need some other way to get the array element's type. Looks like we will need to come up with a new kind of Object. Let's imagine how we would use it: [9aad98abd4a561146ee3048a2b08bc700eed055d]
https://github.com/nyjc-computing/pseudo-9608/blob/9aad98abd4a561146ee3048a2b08bc700eed055d/pseudocode/resolver.py#L78-L79
https://github.com/nyjc-computing/pseudo-9608/blob/9aad98abd4a561146ee3048a2b08bc700eed055d/pseudocode/resolver.py#L183-L194

And finally we design the Array object to make the code work: [faaa0a955b9ff1d6de0dd8c45f4fd1093b373b0d]

ngjunsiang:
Now our program fails silently in the interpreter (because we haven't implemented code there), but we don't get to see any error messages. Let's add some debugging code to catch them in Pseudo: [183665153f8d702209b87f500f5c731589b3726e]

ngjunsiang:
Let's add some code to test array gets: [5bf4d1fda706e92ac8d31e08210f18e67c9d6d1f]
```
https://github.com/nyjc-computing/pseudo-9608/blob/5bf4d1fda706e92ac8d31e08210f18e67c9d6d1f/main.pseudo#L20-L22
```

To make this work, we have to implement array gets and sets. The main problem here is that `name` in the Get Expr is a tuple of Exprs, which will not match the Array keys. We have to turn them into a tuple of values (actually integers):
https://github.com/nyjc-computing/pseudo-9608/blob/e6644f7cd8d17427f27cc44b6ac03b7cdc729f69/pseudocode/interpreter.py#L86-L93
https://github.com/nyjc-computing/pseudo-9608/blob/e6644f7cd8d17427f27cc44b6ac03b7cdc729f69/pseudocode/interpreter.py#L119-L125

They look so similar that it would be criminal not to make a helper here:
f2c892bc8ec9e8fd53d265df5e294ec2680bfd63

ngjunsiang:
A minor fix for missing arguments: [3fc85db3497c4ef13286fa23d170a444918d83ba]

And we are *almost* done. The code works, but we haven't put in the runtime type-checks that we could not do in the resolver.

ngjunsiang:
[0aa04f6c42667872a4ee87c5ab5fbe5589abfb5e]

We make `resolveArray()` and `resolveObj()` helpers. And turns out `value()` isn't being used at all, so let's remove that.

And we're done!

PR #61, opened 2022-05-21
24a Improvements: type annotation

ngjunsiang:
With 80% feature completion, I'm not expecting much changes to the code at this point. That means we can start to lock down the interfaces and app design for more stability.

But first, some inconsequential edits, mostly for readability: [fe90aeddeeeff7a762ce1e50becfe7d2dc1dc811], [37490dc51e6ca09fe7ef194c035037e78d96f326], [95782cd76e0fac09d860041ef88b6e7c310c2ec1], [12da110f17b363670b861037ea4fa17de3313f39], [01cbeb8c56d0538864790b1355ed8c7d76f7b84a]

A fix for a bug introduced during cleanup: [ff2da2a6c5338771d43f7d8744feefe7b9a009dc]

And another helper to help us iterate through combinations of indexes for declaring arrays: [70899ab3574843c38cdcc2691abb4abe80aa2f5b]

Finally, a fix in error reporting: [58283ed08fcdfcc380c0a1f391aef4bd0cddfa9f]

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,10 +10,24 @@ pip install pseudo-9608
 
 ## Usage
 
+### Running psuedocode on a file
+
 ```
 import pseudocode
 
-pseudocode.run('myfile.pseudo')
+pseudocode.runFile('myfile.pseudo')
+```
+
+### Running psuedocode on a string
+
+```
+import pseudocode
+
+code = """
+OUTPUT "Hello World!"
+"""
+
+pseudocode.run(code)
 ```
 
 # Chapters
--- a/pseudocode/__init__.py
+++ b/pseudocode/__init__.py
@@ -1,6 +1,6 @@
 from typing import Iterable, Mapping, Callable as function
 
-import builtin
+from . import builtin
 from .lang import Frame
 from .system import system
 
--- a/pseudocode/__init__.py
+++ b/pseudocode/__init__.py
@@ -1,4 +1,6 @@
 from typing import Iterable, Mapping, Callable as function
+from sys import exc_info
+from traceback import format_exception
 
 from . import builtin
 from .lang import Frame
@@ -13,6 +15,12 @@ from .interpreter import Interpreter
 __version__ = '0.2.1'
 
 
+
+# https://stackoverflow.com/a/66416364/318186
+def printException():
+    etype, value, tb = exc_info()
+    info, error = format_exception(etype, value, tb)[-2:]
+    print(f'Exception in:\n{info}\n{error}')
 
 def error(lines: Iterable, err: builtin.PseudoError) -> None:
     errType = type(err).__name__ + ':'
@@ -61,8 +69,9 @@ class Pseudo:
         except builtin.ParseError as err:
             result['error'] = err
             return result
-        except Exception as err:
-            breakpoint()
+        except Exception:
+            printException()
+            return result
 
         # Resolving
         resolver = Resolver(globalFrame, statements)
@@ -71,8 +80,9 @@ class Pseudo:
         except builtin.LogicError as err:
             result['error'] = err
             return result
-        except Exception as err:
-            breakpoint()
+        except Exception:
+            printException()
+            return result
 
         # Interpreting
         interpreter = Interpreter(globalFrame, statements)
@@ -81,7 +91,7 @@ class Pseudo:
             interpreter.interpret()
         except builtin.RuntimeError as err:
             result['error'] = err
-        except Exception as err:
-            breakpoint()
+        except Exception:
+            printException()
         finally:
             return result
--- a/pseudocode/__init__.py
+++ b/pseudocode/__init__.py
@@ -1,4 +1,6 @@
-from .builtin import ParseError, RuntimeError, LogicError
+from typing import Iterable, Mapping, Callable as function
+
+import builtin
 from .lang import Frame
 from .system import system
 
@@ -12,7 +14,7 @@ __version__ = '0.2.1'
 
 
 
-def error(lines, err):
+def error(lines: Iterable, err: builtin.PseudoError) -> None:
     errType = type(err).__name__ + ':'
     if err.line:
         lineinfo = f"[Line {err.line}]"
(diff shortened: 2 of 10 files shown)


ngjunsiang:
# Removing the Visitor pattern

It turns out after the cleanup that we don't really need the Visitor pattern after all. Oh well, it was helpful for a while, but we've managed to organise things better! So away with the Visitor pattern: [1ce824238f0e9387466bf627e957215edc028754], [468da32531ef1478d56a3527f286e6452b94f2e3]

ngjunsiang:
Okay, back to actual work.

If I have to step away from this project for a few months, when I return I would have a hard time figuring out what's going on. That means any newcomer to this project will be struggling to build a model of how this interpreter works. We can improve that.

There's much documentation to be return, but let's start with something more basic: knowing what types to expect. We'll use Python's support for [type hinting](https://docs.python.org/3/library/typing.html).

# Type hinting

## builtin.py

Our `builtin` is the only exception to this type annotation rule: Since it is going to be imported by all other modules, and shouldn't depend on any of them, it can't actually know anything about how it is going to be used. So no type annotation here. And frankly I'm fine with that, except for the errors: those clearly need to know what they are dealing with. I'll move the errors out later, not now.

## lang.py

We start by defining some basic types used in pseudo:
https://github.com/nyjc-computing/pseudo-9608/blob/a77f14baf1a29eb38cd3cbca6f5d07118393573f/pseudocode/lang.py#L1-L13

We do this so if we ever decide to, say, change Type to something other than `str` (an Enum maybe?) we only need to change it in one place instead of >20.

Some are not annotated here, e.g. File open modes. These are typically used in only one place and not really worth the effort to make a new type for; I might change my mind in future though.

Then we add type annotations to all the method definitions: [Changes: [a77f14baf1a29eb38cd3cbca6f5d07118393573f]]

We use quoted types (e.g. "Expr") for types that are defined as classes within the same module, to avoid "undeclared name" errors.

We'll start

ngjunsiang:
## system.py

Very little to annotate here, thankfully. There will be more. [0c625d3a2500f42f01369ecbd8ab973fed4e1a68]

ngjunsiang:
## __init__.py

Also very little to annotate here: [51370ba010e2c8eddcf566bff5671f804ec8094d]

ngjunsiang:
## scanner.py

Our scanner doesn't need that many types. We define a custom type for Code, which for now is just a dict (i.e. Mapping) but may become its own object in future.

Our scanner functions all return strings, except the `makeToken(0` helper which returns a token. And the `scan(0` entry point returns a 2-ple consisting of a list of tokens and a list of code lines (as `str`).

The changes: [7054c26c63836a407600b5614bde8bfd6391cae3]

ngjunsiang:
Some fixes - I did relative imports wrongly: [aa6134b719e9826fb4021439a51ad9f1f2215cd2]

And then type annotation for the parser: [0fa8b48b8e7ac4e230771b10fecda31de5790eb9]

I decided to import the entire `lang` module instead of importing each separate item in.

This one is a bit more involved. Some helpers either return a token or raise an error, so we use `Optional[]` typing to indicate this. We currently represent our tokens as a list of `Token`, and this is referenced so much that I made a custom type for it instead of repeating `Iterable[lang.Token]` so much.

ngjunsiang:
## resolver.py

The resolver and interpreter are going to be rather more involved, because there are so many resolvers, verifiers, evaluators, and executors.

Let's start by simplifying imports: [954494910e1f5a5c48acd64b78d95f055f67c47e]

And then ... all the type annotations: [f76f65d275eb5d58f27cdad43d316a8b19b60408]

Notable is `rangeProduct()`, which takes in a list of `(start, end)` tuples and returns an iterator:
https://github.com/nyjc-computing/pseudo-9608/blob/f76f65d275eb5d58f27cdad43d316a8b19b60408/pseudocode/resolver.py#L42

Also made a simplification in stringifying types before reporting errors:
https://github.com/nyjc-computing/pseudo-9608/blob/f76f65d275eb5d58f27cdad43d316a8b19b60408/pseudocode/resolver.py#L22-L25

Also the return types from Procedure & Function calls:
https://github.com/nyjc-computing/pseudo-9608/blob/f76f65d275eb5d58f27cdad43d316a8b19b60408/pseudocode/resolver.py#L242-L253
https://github.com/nyjc-computing/pseudo-9608/blob/f76f65d275eb5d58f27cdad43d316a8b19b60408/pseudocode/resolver.py#L255-L270

The resolvers *must* return a type (currently `str`), while the verifiers don't return anything (except `verify()` which optionally returns type from return and call statements).

ngjunsiang:
# interpreter.py

Finally, the interpreter. After what we did with the resolver, everything here follows a similar grain: [d0fd220bf50f797b27484b777926f69448db6eac]

ngjunsiang:
# Static analysis

Our type annotations pay off almost immediately. By installing the [mypy](http://mypy-lang.org/) module, we can run a static analysis on our Python files to look for type inconsistencies with `python -m mypy pseudocode/*.py`.

The inconsistencies are long! Here's a short snippet:

```
pseudocode/lang.py:62: error: Need type annotation for "data" (hint: "data: Dict[<type>, <type>] = ...")
pseudocode/lang.py:65: error: Argument 2 to "setTemplate" of "TypeSystem" has incompatible type "None"; expected "Object"
```

Our type system is behaving inconsistently. Let's fix it: [da5563b4cc36fc391963c3454edcff67dd7a1233]

ngjunsiang:
Some of the surfaced problems will require more than a few lines to fix, so we'll just create new issues for them for now:
- #62 

In more than a few cases, it has forced me to be stricter about expected/accepted types: [c8fc73a2eeb218944fb00b5bffb0b9ae4921eb56]

And helped me catch some variable shadowing: [2c25c7bf709f10565fe5310b5d06ec1b394fbea6]

ngjunsiang:
# Error printing

Meanwhile, because of our `try-except` statements in `__init__.py`, we are no longer getting er4ror reports for *internal* errors (i.e. anything that isn't a `ParseError`, `LogicError`, or `RuntimeError`).

With some help from https://stackoverflow.com/a/66416364/318186, we make our own error printer to help us with debugging:
https://github.com/nyjc-computing/pseudo-9608/blob/6391be6718377a72027f75434a41668ef863320b/pseudocode/__init__.py#L19-L23

And then we can use it to print any non-pseudo errors we encounter: [6391be6718377a72027f75434a41668ef863320b]

ngjunsiang:
# Type Aliasing

Type aliasing proves tricky; when we state `Arg = "Expr"`, Python can't figure out if `Arg` is a literal `str` with value `"Expr"`, or an alias for the Expr type. So we do aliasing the proper way, with `typing.NewType`: [7954c1eb1971f5bc94488c26e8f42861394a917e]

ngjunsiang:
Our parsers, resolvers, and executors are all pretty loose with their type annotations. For instance, `evalUnary()`, instead of expecting a Unary, expects a generic Expr. This makes it hard for `mypy` to do its job. So let's tighten up on those annotations: [68864ac4a323a99cee704d63c4e85e850b75112d]

ngjunsiang:
More typesystem improvements: the aprameter names make it difficult to understand what it is doing, so we do a long of parameter and method renaming: [8fabee0f75bf8dae0995c283d6c51c09c56e634c]

ngjunsiang:
We have a whole class of headaches like `pseudocode/interpreter.py:323: error: "Stmt" has no attribute "rule"` that are going to need quite a big change in our type system and hierarchy. We'll get to that next chapter and really clean things up.

PR #64, opened 2022-05-22
24b Improvements: Decoupling

ngjunsiang:
`pseudocode/interpreter.py:323: error: "Stmt" has no attribute "rule"`

Let's start with this:
https://github.com/nyjc-computing/pseudo-9608/blob/d59a6b052fdc827d3f7b29892aadc55822b003c0/pseudocode/lang.py#L496-L497

This tells Python that the `Stmt` class has a `rule` class attribute, and any class that inherits this should override the `rule` attribute (or risk getting a `NotImplemented` value when trying to access it`.

And now that message goes away when we run `mypy` again.


ngjunsiang:
We should define interfaces like this for any class that gets inherited. Let's look at Expr:
https://github.com/nyjc-computing/pseudo-9608/blob/ab0133c99b49da3732ff2fb5671eb1f006524cfe/pseudocode/lang.py#L358-L377

Not having an `__init__()` makes it easier to implement the subclasses flexibly, while still ensuring that they implement a way to retrieve a token.

Immediate implications: looks like we'll have to decouple Literal from TypedValue and give it its own `__init__()`:
https://github.com/nyjc-computing/pseudo-9608/blob/725147250c9dc693f90cab9d482e5b1f27bc23dc/pseudocode/lang.py#L381-L392

And it now implements its own `token()` method as well. We'll do the same for each Expr subclass; tedious, but has some advantages as we'll see later.

ngjunsiang:
Looking at our expectations of what an Expr is and does:
https://github.com/nyjc-computing/pseudo-9608/blob/7c2d2a743563bd1df1bfb3d5f3d18e7da1250d4b/pseudocode/lang.py#L358-L369

It seems like Name should no longer be considered an Expr: it does not resolve to a Type, and doesn't evaluate to a Value as well (note that names are not one of pseudo's types). So we take out its inheritance from Expr:
https://github.com/nyjc-computing/pseudo-9608/blob/7c2d2a743563bd1df1bfb3d5f3d18e7da1250d4b/pseudocode/lang.py#L396-L407

It would, however, still be useful to retain the token associated with the name, so we'll keep that.

ngjunsiang:
The `makeExpr()` helper was useful while we were still unsure how tokens for error reporting were going to work; we could decide at the point of instantiation which token to use.

Now, many chapters later, this did not come to pass; for the most part each Expr will use a specific part of itself as the associated token. For example, Declare Exprs are always associated with their name token:
https://github.com/nyjc-computing/pseudo-9608/blob/7c2d2a743563bd1df1bfb3d5f3d18e7da1250d4b/pseudocode/parser.py#L348-L353

It no longer makes sense to be passing tokens to Expr at init time then. Let's update Declare: [425544d6b06976f6429252dff70e043ac6b50d23]
https://github.com/nyjc-computing/pseudo-9608/blob/425544d6b06976f6429252dff70e043ac6b50d23/pseudocode/lang.py#L411-L424

This is a breaking change: names are no longer plain strings, but Name objects with associated tokens. this will involve trickle-down changes in the parser, resolver, and interpreter.

ngjunsiang:
Similar changes for Unary and Binary: [7eec4033c440d4606b167b94ffda57c025a0b49f]

They also lose the superclass init. Unfortunately, their operators are passed as functions: we generally want to avoid Exprs knowing about how tokens work; we can pass tokens around but try not to extract information from them manually.

This means we will have to instantiate Unary and Binary with a `token=` keyword argument.

ngjunsiang:
And finally, our most complex Exprs. Call no longer needs to hold on to a token: [9682e6ee57acfe3a4d18908b73f6221046ef320e]
https://github.com/nyjc-computing/pseudo-9608/blob/9682e6ee57acfe3a4d18908b73f6221046ef320e/pseudocode/lang.py#L484-L496
https://github.com/nyjc-computing/pseudo-9608/blob/9682e6ee57acfe3a4d18908b73f6221046ef320e/pseudocode/lang.py#L500-L511

ngjunsiang:
A few fixes in the way we use Name: [06b3e87e60cb2514e7a350c84fcd02f229cecbc2]

This is mighty confusing. I'm going to redo this part when I figure out a way to pass names and tokens around more cleanly.

`makeExpr()` makes static type checking difficult. So let's work towards eliminating it, by replacing all `makeExpr()` calls with their appropriate class instantiations: [904a050e969b5b3a7da7059fbb1479e37e5b9156]

And then remove it: [622be6d04ae7f23c9ff10407bad65a17d24ea464]

ngjunsiang:
More fixes:
- I misunderstood the use of `NewType`; Arg is unnecessary and Index is just typed as a tuple of Exprs or tuple of ints: [ae5fd65a5512bc37dbcf1c02eb0afea77557866b]
- Grouping the ExprStmts together for more consistent typing: [4793f4a4ee5812a999c6971e65683d01891124a1]
- `value()` is proving tricky to type properly; this is a huge function with many possible return types. I'll revisit this in a later refactoring chapter: [e96f1a91451df4f49735fab6b3a39218226a5984]

ngjunsiang:
Next, typing fixes for various statement types:
- FileAction: [994923200baea3ed51a16821f679adf269ea631d], [bf6ab77917b6cc269df98cc230cb2d26ad9815c6]
- Conditional: [45512ace67fc78570e5db64377b3eaa15ca03427]
- Assign, where we finally get rid of the `name`attribute: [995e36fe40f00e00242d0a8a679c556fed549677]
- Input (technically not typing fix, but name handling needs some fixing here): [e34e1a374906a3b521f9f2c80fa8b0f5d6c051e2]
- Loop: [0f5e970bb4c87186b38b1f585bbe5349f422d7c6], [236a0f42401e28801a848abc2a735e2ecf040342]
- Procedure/Function: [11dc370cb78378cd3994057ca49b4180811e1c8f]

ngjunsiang:
Then some renaming for better readability: [98383cd2c8342df9e6ef64f354b8ed57553139ee]

Instead of Value, we now use PseudoValue to refer to pseudo's special value types (Callable, File, Array, Object, ...)

Value, on the other hand, refers to any value which can be represented in a TypedValue. So this includes PsuedoValues, bool, int, float, str.

ngjunsiang:
What follows is a lot of mole-whacking for typing inconsistencies, which are ... too numerous to detail here. I simply aimed to have working code for now, and next chapter we will go about organising things in a clearer fashion that will make it easier for us to reason about the code.

On the plate are:
- clear parse/resolve/execute dispatch and return types
- clearer object hierarchy

PR #65, opened 2022-05-22
24c Improvements: Type Relationship

ngjunsiang:
We kick things off with a minor but much needed clarification on how to use names:
https://github.com/nyjc-computing/pseudo-9608/blob/79cdd1d136dde2dabe92521b4a2687db4a347597/pseudocode/lang.py#L47-L62

Name gains a `__str__()` dunder now, so we can call `str()` on it when we need it in string form: [47c74f52449f8ada3b668aef87c47a65821a5e5a], [3cea52caf16aa1daf9c00b4eed5566042083b3a3]


ngjunsiang:
Before we come up with an object hierarchy, there are a couple of issues we need to think through:
1. The parser represents everything with Exprs and Stmts. Should the resolver evaluate these for the interpreter, or leave them in raw form?
2. Just how many different kinds of objects are there?

Here's a quick mind-dump of all the different things we deal with in this pseudocode interpreter:
- Objects (TYPE), Arrays, Frames, TypeSystems
- Callables, Files
- Exprs (including Literal), Stmts
- Names (any valid but undeclared word representing a stored value) and Types ( built-in types or previously declared type, representing a stored template)
- Indexes (sequence of 1 or more integers, used in Arrays)
- Args (sequence of Exprs) and Params (sequence of TypedValues that take in values from Args)
- TypedValues (Type-Value pair) and TypeTemplates(TypedValue that cannot be assigned to, but can be cloned)

And some of the ways they are related:
- An Object maps Names to TypedValues
- A Frame is an Object with an optional outer Frame
- An Array maps Indexes to TypedValues
- A TypeSystem maps Types to TypeTemplates
- A Callable contains a Frame, and when called, assigns evaluated Args to Params before executing
- A Declare denotes a Name-Type pair to be declared in a Frame/Object
- ...

etcetcetc

ngjunsiang:
Here's an initial draft of our reworked types:

https://github.com/nyjc-computing/pseudo-9608/blob/bd16193abb2b0ab91091ce4a4438f06c41f588a8/pseudocode/lang.py#L5-L17

Why are the last 2 commented out? Because we are going to try to eliminate them!

ngjunsiang:
Objects, Arrays, Frames, TypeSystems, ... one thing they all share in common is that they map keys (a NameKey or IndexKey) to values (a TypedValue or TypeTemplate). They don't share any attributes or method implementations in common, so it doesn't make sense to have them inherit from a common parent (and we'll extend this philosophy to other classes later). Let's instead define a [Protocol](https://peps.python.org/pep-0544/) for them:
https://github.com/nyjc-computing/pseudo-9608/blob/f47ddc1bbd8fc873ffae380d90e7adc15e41a8c8/pseudocode/lang.py#L65-L76

All it takes for a class to qualify as a PseudoMap is that it implements the `has()` method. If we have a function or method that is going to use `has()` to check for the existence of a key, we can use PseudoMap to represent this type without knowing if it will be an Object, Array, Frame, or TypeSystem.

ngjunsiang:
Next, we differentiate TypeTemplates from TypedValues:
https://github.com/nyjc-computing/pseudo-9608/blob/f9e274f579cd308683356f3062dea368083c2fa7/pseudocode/lang.py#L98-L121
https://github.com/nyjc-computing/pseudo-9608/blob/f9e274f579cd308683356f3062dea368083c2fa7/pseudocode/lang.py#L80-L94

This makes it clear that a TypeTemplate can be cloned to create a TypedValue, but not vice-versa. We will use TypeTemplates in TypeSystems, and TypedValues in Objects, Arrays, and Frames.

ngjunsiang:
Why don't we implement PseudoMap as a base class instead of a protocol? We could have Object, Array, etc inherit PseudoMap, then override the `has()` abstractmethod and implement other methods.

Unfortunately, that's not going to work for `mypy`, because [it follows the Liskov Substitution Principle](https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases) . In short, everywhere and every way that PaeudoMap is used, its subclasses must support too.

If PseudoMap.has() takes a key of any type, then Object, Array, etc must be able to handle that. With Object taking only keys of NameKey type ans Array taking only keys of IndexKey type, clearly that can't happen.

So a protocol is an easier way to describe this requirement.

ngjunsiang:
Time to rename types for consistency: [39b8e50a7a0901d227dee7bce4e8bd1f7588297e]

With our new TypeTemplate, we can now make TypeSystem easier to understand:
https://github.com/nyjc-computing/pseudo-9608/blob/195e63053ce0d1c9e21a5d649f35fc44fd31abe4/pseudocode/lang.py#L127-L171

The `clone()` method of the TypeTemplate really makes the work of the TypeSystem much easier.

ngjunsiang:
Now for the Object:
https://github.com/nyjc-computing/pseudo-9608/blob/fb5295542b90c13d3a9d5212072eea05a613caf0/pseudocode/lang.py#L183-L238

Our Object doesn't need a `clone()` anymore, so let's remove that. It also doesn't need to handle TypeValues during declaration; that is now taken up by the TypeSystem.

ngjunsiang:
And now the Frame, which has minimal changes:
https://github.com/nyjc-computing/pseudo-9608/blob/a35a835f2528e5bde79ae1c8e378e5396a035b7a/pseudocode/lang.py#L242-L278

Just needed to add type annotations and update the docstring mostly.

ngjunsiang:
Pretty big changes to Array:
https://github.com/nyjc-computing/pseudo-9608/blob/1b974029430b20dffe2e6499f39990e9d18e84e5/pseudocode/lang.py#L283-L374

Since arrays are statically allocated, I decided we might as well do it in the init, which requires us to pass in the ranges to the init. So we also port over the `rangeProduct()` helper (which used to be in `resolver.py`) as a static method of the class.

Array also gains some attributes/properties describing it, in case we need them. And loses `declare()` since we don't need it. Separating it from Object also means we can do stricter type annotations.

ngjunsiang:
Our Callables are fine, no real need to touch them for now.

ngjunsiang:
Type-wise, File is a mess:
https://github.com/nyjc-computing/pseudo-9608/blob/1b974029430b20dffe2e6499f39990e9d18e84e5/pseudocode/lang.py#L449-L474

Depending on the mode passed to it, the meaning of `name` and `data` changes. No wonder so many `mypy` static typing errors are coming from it!

Let's break it up into multiple classes, one for each mode.

ngjunsiang:
There's also not much to be done to the Expr base class. We could specify some required attributes and methods, using [`@abstractmethod`](https://docs.python.org/3.8/library/abc.html#abc.abstractmethod) for the latter: [b6edc788375408f5b270d2f945eaf73f5d341c93]

ngjunsiang:
Lots of tiny fixes for Expr and Stmt subclasses, I won't detail those; we are essentially applying the type relationship established earlier: [20fc31c6655395e17723178e460714b2e14b4be9]

And then a couple of big changes.

ngjunsiang:
## De-confusing FileAction

Our poor, confused FileAction has been trying to hold 4 types of File statements all this time:
https://github.com/nyjc-computing/pseudo-9608/blob/696a1365add62400edd11d64624186694a53058d/pseudocode/lang.py#L738-L752

This worked fine at the time, when we were trying to avoid adding too much complexity at once. Now, with most of the interpreter in place, and with us trying to annotate types so that collaboration will be easier, it's time to break them up: [6eab69426baded4e914ef662d257a42d4668ca39]
https://github.com/nyjc-computing/pseudo-9608/blob/6eab69426baded4e914ef662d257a42d4668ca39/pseudocode/lang.py#L738-L782

Yep, it's a lot more code, but also much easier to annotate types for, and easier to understand.

ngjunsiang:
## Differentiating Gets

Earlier, we decided that a Get accepts a `frame` and a `name`. This makes sense for variable names, but our current interpreter also uses Gets for array-indexing (i.e. `Name[0]`) and attribute-getting (i.e. `Name.Attribute`). Both will not pass static type-checking if we only accept Frame.

So it's time to break up Get as well: [7c70edba71d9f7d5c5e8d849e20ab4b3520e424d]
https://github.com/nyjc-computing/pseudo-9608/blob/7c70edba71d9f7d5c5e8d849e20ab4b3520e424d/pseudocode/lang.py#L593-L634

Much more code, but much clearer! We don't have to balance in our heads the differences between Array, Object, and Frame, and there will be less need for `if`-checking.

ngjunsiang:
# Typing Assign

What about our poor Assign though?
https://github.com/nyjc-computing/pseudo-9608/blob/7c70edba71d9f7d5c5e8d849e20ab4b3520e424d/pseudocode/lang.py#L540-L548

It's still expecting a Get for its assignee ... but Get doesn't exist any more. We could reasonably assign a value to a variable name, array index, or object attribute, so how should we type this?

The tedious way is to define another Protocol for a platonic ideal of a Get. The simpler way is just to `Union` the three types:
https://github.com/nyjc-computing/pseudo-9608/blob/4b85bad8241dff5a5b10b643d56dafef0dcbdfbe/pseudocode/lang.py#L540-L551

Wait, didn't I say earlier that `Union`ing multiple types is a red flag? Yes, but only if they are very different types. Here, all three types are Gets, i.e. retrieving a TypedValue from a map using a key, but for different kinds of maps and different kinds of keys. The resolver and interpreter for assign can still reasonably deal with them (with more resolvers/interpreters for the various Gets).

ngjunsiang:
# Bugfixing

`mypy` still throws the following errors:
```
pseudocode/lang.py:123: error: "Object" has no attribute "copy"
pseudocode/lang.py:166: error: Unsupported target for indexed assignment ("Mapping[str, TypedValue]")
pseudocode/lang.py:172: error: "TypedValue" has no attribute "clone"
pseudocode/lang.py:227: error: Incompatible types in assignment (expression has type "Optional[Object]", target has type "TypedValue")
pseudocode/lang.py:322: error: Value expression in dictionary comprehension has incompatible type "Optional[Object]"; expected type "TypedValue"
pseudocode/lang.py:322: error: Argument 1 to "cloneType" of "TypeSystem" has incompatible type "Type[type]"; expected "str"
pseudocode/lang.py:354: error: Argument 1 to "len" has incompatible type "Iterable[Tuple[int, int]]"; expected "Sized"
pseudocode/lang.py:357: error: Missing return statement
pseudocode/lang.py:424: error: Variable "pseudocode.lang.Param" is not valid as a type
pseudocode/lang.py:424: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
pseudocode/lang.py:657: error: "Stmt" has no attribute "__slots__"
pseudocode/lang.py:738: error: Variable "pseudocode.lang.Param" is not valid as a type
pseudocode/lang.py:738: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
```

Let's fix them.

ngjunsiang:
First, let's fix the recursive cloning of TypeTemplates: [3188480a008e6d76848686801f4512a2798757b7]

Previously, if the TypeTemplate contained an object, we cloned it. With our stricter type system, we don't expect to clone objects, we expect to clone templates. So it's time to make an ObjectTemplate:
https://github.com/nyjc-computing/pseudo-9608/blob/946625bcc3258b4478b5d03497a97df22f525ad7/pseudocode/lang.py#L128-L156

This lets users declare nested TYPEs, but if the user declares an object like
```
TYPE
    DECLARE Data : ARRAY[1:5] OF INTEGER
ENDTYPE
```

We don't have an ArrayTemplate, so this isn't going to work. Let's mark it as an unsupported feature and add it to issues: #66

ngjunsiang:
On to the remaining issues.

```
pseudocode/lang.py:123: error: "Object" has no attribute "copy"
pseudocode/lang.py:166: error: Unsupported target for indexed assignment ("Mapping[str, TypedValue]")
pseudocode/lang.py:172: error: "TypedValue" has no attribute "clone"
pseudocode/lang.py:227: error: Incompatible types in assignment (expression has type "Optional[Object]", target has type "TypedValue")
pseudocode/lang.py:322: error: Value expression in dictionary comprehension has incompatible type "Optional[Object]"; expected type "TypedValue"
pseudocode/lang.py:322: error: Argument 1 to "cloneType" of "TypeSystem" has incompatible type "Type[type]"; expected "str"
```
All fixed in [6b7f8130f5311996b0385c460e2c02592b8de8c7], just a whole bunch of un-updated types.

Python also does not expect a Mapping to be assigned to; we need a MutableMapping for that.

```
pseudocode/lang.py:424: error: Variable "pseudocode.lang.Param" is not valid as a type
pseudocode/lang.py:424: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
pseudocode/lang.py:738: error: Variable "pseudocode.lang.Param" is not valid as a type
pseudocode/lang.py:738: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
```
Turns out newer versions of `mypy` do not treat `Param = "TypedValue"` to mean that Param is equivalent to a TypedValue; we need to declare a TypeAlias for that. But TypeAliases are not yet natively supported in Python 3.8, which I am using. I googled and found that `Union` can be used as a workaround for now: [113cc19504067c6d9fac33311368bd3ff7d34143]

```
pseudocode/lang.py:657: error: "Stmt" has no attribute "__slots__"
```
Oh? Now it does: [fdb8cd0b6cb1f929d29bd615d5c19dbc3fc1d6ed]

```
pseudocode/lang.py:354: error: Argument 1 to "len" has incompatible type "Iterable[Tuple[int, int]]"; expected "Sized"
pseudocode/lang.py:357: error: Missing return statement
```
Turns out Iterators need not necessarily have a known size, so we can't use `len()` on them. We need to use the Collection type, which is expected to have a known size: [a044111404bbfcd24f1349f205e9959f889c0f7d]

ngjunsiang:
```
$ python -m mypy --follow-imports=skip pseudocode/lang.py
Success: no issues found in 1 source file
```
Ah, satisfying 😎

PR #67, opened 2022-05-25
24d Improvements: Parser

ngjunsiang:
On to the parser. By itself it doesn't have any type problems, but we do need to update it so as to reflect the changes in `lang.py`.

First off, let's fix a leftover type issue in lang`: [157874740dbacb99e49fc1b07f5b8a4ed7e716fd]

And begin with an easy fix: We decided that names would be parsed as a Name Expr, and replaced with a Get Expr in the resolver. Turns out this applies in two places, where we use `name()` instead of `identifier()`. Let's change those calls to `identifier()`, and comment out `name()`: [157874740dbacb99e49fc1b07f5b8a4ed7e716fd]


ngjunsiang:
# Using the Get family

First up, let's use our newly created Get family of Exprs:
https://github.com/nyjc-computing/pseudo-9608/blob/3c0979456aa1f9ff3dd4f6b49eb95db4998baafc/pseudocode/parser.py#L111-L120

For now we only support literal indexes, i.e. `[1,2,...]`. No using of expressions yet, i.e. `[Num1, Num2, ...]` is not supported. This can be a future feature: #68

We better type-check in `literal()` then:
https://github.com/nyjc-computing/pseudo-9608/blob/3c0979456aa1f9ff3dd4f6b49eb95db4998baafc/pseudocode/parser.py#L86-L88

Not forgetting to rename `arrayExpr()` to `indwexExpr()`:
https://github.com/nyjc-computing/pseudo-9608/blob/3c0979456aa1f9ff3dd4f6b49eb95db4998baafc/pseudocode/parser.py#L139-L140

ngjunsiang:
For this to work, we need to break up TYPES in `builtin.py` even further to include a LITERAL set:
https://github.com/nyjc-computing/pseudo-9608/blob/3c0979456aa1f9ff3dd4f6b49eb95db4998baafc/pseudocode/builtin.py#L100-L108

ngjunsiang:
Gotta fix the return types too:
https://github.com/nyjc-computing/pseudo-9608/blob/b57ce9729af2e729e11e7429ad278894256df906/pseudocode/parser.py#L108
https://github.com/nyjc-computing/pseudo-9608/blob/b57ce9729af2e729e11e7429ad278894256df906/pseudocode/parser.py#L112-L113
Since we defined an IndexExpr type, we can use that (line 113) too.

I also took the chance to simplify `callExpr()` a little, minor change really:
https://github.com/nyjc-computing/pseudo-9608/blob/b57ce9729af2e729e11e7429ad278894256df906/pseudocode/parser.py#L99-L106

We don't actually have a Get type anymore, but we can recreate it:
https://github.com/nyjc-computing/pseudo-9608/blob/b57ce9729af2e729e11e7429ad278894256df906/pseudocode/lang.py#L22

We need this because GetIndex and GetAttr still need a Get for an Array or Object ... which can plausibly be assigned to a variable name, a parent object's attribute, or even a parent array's index. (Of course, we would not expect arrays to be retrieved from another array since nested array declarations are not supported by 9608 pseudocode.)

ngjunsiang:
Next, using our new *File Stmts: [32152e0a2289434b38ca8619b39cc38733f6ac43]

Nothing unexpected here, at least not until we get to the resolvers and executors.

ngjunsiang:
Minor cleaning up: [0a5a18cd797bce3d96cc73ebcbcb5f6796f737fd], [e9547865d4e45d370d9e170bb9d4e89ce8e5c10b]

ngjunsiang:
While it's not part of the parser, we do need to update our scanner as well: [c420cfb5795731e81a5503961a6aa9a0ff764617]
Mostly minor type updates from whatever we changed in `lang.py`, except for one line:
https://github.com/nyjc-computing/pseudo-9608/blob/c420cfb5795731e81a5503961a6aa9a0ff764617/pseudocode/scanner.py#L73

Here I decided to annotate the return type more explicitly. Return types are ideally strict, so they can match different expected param types.

ngjunsiang:
Another round of checking, this time with multiple files to check for cross-module type issues.

More minor fixes: [816f06c64e70226ebb88fd027856954705f59770]

This one involves adding yet another type to represent possible values that a Token can take:
https://github.com/nyjc-computing/pseudo-9608/blob/816f06c64e70226ebb88fd027856954705f59770/pseudocode/lang.py#L11

This feels ugly, I'll come back to look at it again in future chapters.

ngjunsiang:
The rest of the scanner's type problems go away with a Code class: [a557ba88a35e61324bd0fadbbb58981a67362cbf]
https://github.com/nyjc-computing/pseudo-9608/blob/a557ba88a35e61324bd0fadbbb58981a67362cbf/pseudocode/scanner.py#L70-L94

This lets us annotate the type for each attribute, which we can't do with dictionary values. We can even encapsulate the logic for incrementing line numbers in a `nextLine()` method.

ngjunsiang:
Now back to cross-module type issues for the parser.

ngjunsiang:
The remaining type errors are ... thorny issues that will probably require more breaking up. Let's categorise them:

```
pseudocode/parser.py:93: error: Argument 2 to "Unary" has incompatible type "Union[Name, Expr]"; expected "Expr"
pseudocode/parser.py:150: error: Argument 1 to "Binary" has incompatible type "Union[Name, Expr]"; expected "Expr"
pseudocode/parser.py:150: error: Argument 3 to "Binary" has incompatible type "Union[Name, Expr]"; expected "Expr"
pseudocode/parser.py:151: error: Incompatible return value type (got "Union[Name, Expr]", expected "Binary")
pseudocode/parser.py:283: error: Argument 2 to "Conditional" has incompatible type "Union[Name, Expr]"; expected "Expr"
pseudocode/parser.py:329: error: Incompatible types in assignment (expression has type "Union[Name, Expr]", variable has type "Expr")
pseudocode/parser.py:336: error: Argument 3 to "Binary" has incompatible type "Union[Name, Expr]"; expected "Expr"
pseudocode/parser.py:369: error: Incompatible types in assignment (expression has type "Union[Name, Expr]", variable has type "Call")
pseudocode/parser.py:407: error: Argument 2 to "OpenFile" has incompatible type "Union[Name, Expr]"; expected "Expr"
pseudocode/parser.py:414: error: Argument 2 to "ReadFile" has incompatible type "Union[Name, Expr]"; expected "Expr"
pseudocode/parser.py:421: error: Argument 2 to "WriteFile" has incompatible type "Union[Name, Expr]"; expected "Expr"
pseudocode/parser.py:426: error: Argument 2 to "CloseFile" has incompatible type "Union[Name, Expr]"; expected "Expr"
```
These come about because `value()` can return either a Name or one of various Exprs. But the parsers that use `value()` all expect an Expr.

```
pseudocode/parser.py:102: error: Argument 1 to "Call" has incompatible type "Name"; expected "GetName"
pseudocode/parser.py:106: error: Argument 1 to "GetAttr" has incompatible type "Name"; expected "Object"
pseudocode/parser.py:113: error: Argument 1 to "GetIndex" has incompatible type "Name"; expected "Array"
pseudocode/parser.py:133: error: Incompatible types in assignment (expression has type "GetIndex", variable has type "Name")
pseudocode/parser.py:136: error: Incompatible types in assignment (expression has type "Call", variable has type "Name")
pseudocode/parser.py:139: error: Incompatible types in assignment (expression has type "GetAttr", variable has type "Name")
pseudocode/parser.py:197: error: Incompatible types in assignment (expression has type "GetIndex", variable has type "Name")
pseudocode/parser.py:200: error: Incompatible types in assignment (expression has type "GetAttr", variable has type "Name")
pseudocode/parser.py:203: error: Argument 1 to "Assign" has incompatible type "Name"; expected "Union[GetName, GetAttr, GetIndex]"
```
These are still coming from the fact that the parser cannot insert a frame yet, and we have yet to come up with a way to represent Gets-without-Frame from Gets-with-Frame

```
pseudocode/parser.py:88: error: Argument 2 to "Literal" has incompatible type "Union[Union[bool, int, float, str], Callable[..., Any], object]"; expected "Union[bool, int, float, str]"
pseudocode/parser.py:93: error: Argument 1 to "Unary" has incompatible type "Union[Union[bool, int, float, str], Callable[..., Any], object]"; expected "Callable[..., Any]"
pseudocode/parser.py:150: error: Argument 2 to "Binary" has incompatible type "Union[Union[bool, int, float, str], Callable[..., Any], object]"; expected "Callable[..., Any]"
pseudocode/parser.py:158: error: Argument 2 to "Binary" has incompatible type "Union[Union[bool, int, float, str], Callable[..., Any], object]"; expected "Callable[..., Any]"
pseudocode/parser.py:167: error: Argument 2 to "Binary" has incompatible type "Union[Union[bool, int, float, str], Callable[..., Any], object]"; expected "Callable[..., Any]"
pseudocode/parser.py:176: error: Argument 2 to "Binary" has incompatible type "Union[Union[bool, int, float, str], Callable[..., Any], object]"; expected "Callable[..., Any]"
pseudocode/parser.py:185: error: Argument 2 to "Binary" has incompatible type "Union[Union[bool, int, float, str], Callable[..., Any], object]"; expected "Callable[..., Any]"
```
These are token-related issues: token `value`s can taking on a wide variety of types, while the parser functions expect a narrower range of types.

```
pseudocode/parser.py:244: error: Incompatible types in assignment (expression has type "str", target has type "List[Tuple[Any, ...]]")
```
`mypy` really does not like dicts with heterogeneous value types

ngjunsiang:
# Back to GetName ...?

Okay, looks like we do have to parse Exprs instead of Names after all, to please the static typing compiler deities ... let's do that then, but without the `frame` argument:
https://github.com/nyjc-computing/pseudo-9608/blob/002fbd6b44a3bf7690c04294b8cff2c749f8c180/pseudocode/lang.py#L632-L641

This is essentially a GetName without a `frame`. Easier to understand without having to check for NULL frames!

Now, instead of expecting a Name, we can get our expressions to expect an UnresolvedName. Better still, expect a Get (which allows UnresolvedName, GetName, GetAttr, or GetIndex) so we can replace the UnresolvedName with a GetName when resolving: [0cfe6954758eef108b294e04b525de2e92d84e19]

ngjunsiang:
# Reimagining `value()`

Looks like `value()` is a big source of type errors, and no wonder: it has so many responsibilities! It works fine though, but if I come back to this after a long hiatus I would admittedly have a hard time understanding again what it does.

So instead, let's break up its scope of responsibility into multiple functions:
- `unary()` parses and returns a Unary
- `grouping()` parses and returns an Expr between parentheses
- `literal()` parses a Literal
- `identifier()` parses and returns a UnresolvedName
- `name()` parses and returns a Get
- `indexExpr()` parses and returns a GetIndex
- `callExpr()` parses and returns a Call
- `attrExpr()` parses and returns a GetAttr

What can `value()` do then, if it is to have a more predictable return type? Why, we can have it return a *function* then! So instead of `value()` *figuring out* how to parse, and then *parsing* the tokens, now we break up the parsing strategy decision and the parsing action: [df6eb48315e7bdf49d861c877fd0ac2a03a7235e]
https://github.com/nyjc-computing/pseudo-9608/blob/df6eb48315e7bdf49d861c877fd0ac2a03a7235e/pseudocode/parser.py#L97-L101
https://github.com/nyjc-computing/pseudo-9608/blob/df6eb48315e7bdf49d861c877fd0ac2a03a7235e/pseudocode/parser.py#L123-L135
These are parsers for specific types of Exprs

https://github.com/nyjc-computing/pseudo-9608/blob/df6eb48315e7bdf49d861c877fd0ac2a03a7235e/pseudocode/parser.py#L137-L149
And then `parser()` decides which parser to use, based on what the next token is.

ngjunsiang:
Then we replace `value()` with `parser()`: [993e2c4d94ccda03042219b27b4bbbb842813539]

ngjunsiang:
I'm still not quite happy with the type names, still going to fiddle more with them: [ee04afeae4cc11ac9e3fbdf9ac32f70f95b3fb96], [8dcf6c66ceda670245e26b47f775fcc1bde5848a], [0a476cc713e2df2a39d0957aa6cf02b0c84670bb]

Input's `name` argument is actually an Expr that evaluates to a filename, so let's modify that: [491523c3f790229835da77ecaff2a13f46b5d162]

## Get/Index/Attr chaining

Thinking about it carefully, since 9608 pseudocode does not support first-class functions, there is no way an attribute or index can point to a Callable; these can only be declared in a Frame (global frame for declared functions, system frame for Builtins). Let's tweak the `name()` parser to reflect this: [984592c8e2b87d852be0a4b6cf2803d32e00bdeb]

## Token value

Finally I am going to cheap out on the Token.value issue by just treating it as Any: [625f3eaf02ec63cad116bbb2ac1ee092955287a6]

ngjunsiang:
# Resolver type issues

We could spend more time on the parser, but I want to get back to working code first. Let's propagate our Expr changes to the resolver as well.

A first round of type refinements and fixes: [88f5cd9f2ca4dbb08d9103f7e4af6765ed610abd]

Since we are no longer using a NULL frame in a Get Expr, and parsing names as UnresolvedNames instead, let's add code in our Stmt verifiers to resolve them to GetNames: [97e540f8e25e4b1762fd24e87cd70f4e683d59a3]

I have also created resolvers for each subtype of Get:
https://github.com/nyjc-computing/pseudo-9608/blob/97e540f8e25e4b1762fd24e87cd70f4e683d59a3/pseudocode/resolver.py#L185-L229

`resolveGet()` is still sticking around, but delegates the work to the Get subtype resolvers:
https://github.com/nyjc-computing/pseudo-9608/blob/97e540f8e25e4b1762fd24e87cd70f4e683d59a3/pseudocode/resolver.py#L231-L239

ngjunsiang:
On to procedures, functions, and their calls. Some cleanup: [36a55825f2a779dc516fb0a9f0d0f51675e79acc]

Statement verification has been done in the verification functions, and does not need to be repeated in `resolveCall()`.

It also doesn't make sense to be checking procedure statements in `resolveCall()`, not sure why I didn't catch this earlier. Should be done in `verifyProcedure()`: [7172c3979b7725c7fe0e64be540792050258b584]

ngjunsiang:
`resolveCall()`'s name is pretty misleading since all it does is check the args and params, so let's rename it: [e7a1051cdf0c0bc41cab0fc4702f493cdd9b2828]

And take the chance to remove an unused arg as well.

Also feels silly to resolve a callable twice, so let's simplify `resolveArgsParams()`: [7055c38dccf9c494013d2abcec5765cd0b535c70]

Now it really does what it says on the tin.

ngjunsiang:
This chapter is dragging on really long, and it's spilling over into the resolver while we're supposed to be fixing the parser and that is done!

I'll just do one more thing for the resolver before we turn it into its own chapter.

# Eliminating Stmt.rule

With a more complete type hierarchy, we can work towards eliminating Stmt.rule, which makes things a bit more confusing. We'll simply have one type of Stmt for each rule.

Let's start by reworking `verify()` to check the Stmt type instead of checking the rule: [65dab3db21ff759d0749c5f333c404f6645567a7]

In the process, we took apart `verifyExprStmt()` and `verifyFile`. They'll probably reappear before long, don't worry.

ngjunsiang:
https://github.com/nyjc-computing/pseudo-9608/blob/04fecbdb3432db9bfa9d65342ce7c5529b429210/pseudocode/lang.py#L778-L780
https://github.com/nyjc-computing/pseudo-9608/blob/04fecbdb3432db9bfa9d65342ce7c5529b429210/pseudocode/lang.py#L818-L820

Not very much going on here, just adding more subclasses without code. These subclasses replace the `rule` attribute.

And then we use them in `verify()`: [07d7c388490414ad357ecc2bdfd6caeb7745e22a]

Great! We eliminated all use of `rule` in the resolver. (So many things are going to break in the interpreter ...)

ngjunsiang:
But it looks ugly! Let's clean it up a bit. Starting with something small - an `isReturn()` helper: [5197e60e93f4bf784e7051a42bd997d7f8a4f0f4]

All that work of resolving UnresolvedName is adding lots of code bloat. Let's make `resolveName()` help with that too: [6ae86a1b67e7c5d3caa0cb80d4d54a9abd27cd7f]

ngjunsiang:
Now we can remove the `rule` attribute: [daaa26d9210cf6880bc21aa596da696db3fc77ed]

ngjunsiang:
And make the necessary changes for Stmt detection: [1b42063c43d3b0e56b322726dd4ab1dcee77837d], [971b995420d51b1370d58f17477b6949bce2a38c]

Okay, stopping here. For once, I'm going to break the always-keep-code-working rule and continue the changes to the resolver in the next chapter.

PR #69, opened 2022-05-27
24e Improvements: Resolver

ngjunsiang:
continuing where we left off from https://github.com/nyjc-computing/pseudo-9608/pull/67, let's clean up something we left out in the parser: [f32db2fd76c8c65658f470958dad6f873a5f542f], [ca80e95e42c62913f302594753a57f88b958b73e]

### DIFF
--- a/pseudocode/lang.py
+++ b/pseudocode/lang.py
@@ -800,9 +800,27 @@ class Loop(Stmt):
         self.cond = cond
         self.stmts = stmts
 
-class While(Loop): ...
+class While(Loop):
+    def __init__(
+        self,
+        init: Optional["Stmt"],
+        cond: "Expr",
+        stmts: Iterable["Stmt"],
+    ) -> None:
+        self.init = init
+        self.cond = cond
+        self.stmts = stmts
 
-class Repeat(Loop): ...
+class Repeat(Loop):
+    def __init__(
+        self,
+        init: None,
+        cond: "Expr",
+        stmts: Iterable["Stmt"],
+    ) -> None:
+        self.init = init
+        self.cond = cond
+        self.stmts = stmts
 
 
 
@@ -811,7 +829,7 @@ class ProcFunc(Stmt):
     def __init__(
         self,
         name: GetExpr,
-        passby: str,
+        passby: Literal['BYVALUE', 'BYREF'],
         params: Iterable[Declare],
         stmts: Iterable["Stmt"],
         returnType: Type,
--- a/pseudocode/lang.py
+++ b/pseudocode/lang.py
@@ -1,4 +1,4 @@
-from typing import Optional, Union, Protocol, TypedDict
+from typing import Optional, Union, TypedDict
 from typing import Iterable, Iterator, Mapping, MutableMapping, Collection
 from typing import Literal as LiteralType, Tuple, List
 from typing import Callable as function, TextIO
@@ -26,8 +26,6 @@ NameKeyExpr = Union["UnresolvedName", "GetName"]
 GetExpr = Union["UnresolvedName", "GetName", "GetAttr", "GetIndex"]
 # NameExprs start with a name
 NameExpr = Union[GetExpr, "Call"]
-# Rule = str  # Stmt rules
-# FileData = Optional[Union["Expr", str]]
 class TypeMetadata(TypedDict, total=False):
     """The metadata dict passed to an Array declaration"""
     size: Tuple[IndexRange, ...]
@@ -75,21 +73,6 @@ class Name:
 
     def token(self) -> "Token":
         return self._token
-
-
-
-class PseudoMap(Protocol):
-    """
-    Represents a mapping of keys to values used in pseudo.
-
-    Methods
-    -------
-    has(key)
-        returns True if the key exists in the map
-    """
-    @abstractmethod
-    def has(self, key) -> bool:
-        raise NotImplementedError
 
 
 
--- a/pseudocode/lang.py
+++ b/pseudocode/lang.py
@@ -14,7 +14,7 @@ IndexKey = Tuple[int, ...]  # Key for Array
 IndexExpr = Tuple["Literal", ...]  # Array indexes
 IndexRange = Tuple["Literal", "Literal"]  # Array ranges (declared)
 Passby = LiteralType['BYREF', 'BYVALUE']
-Args = Iterable["Expr"]  # Callable args
+Args = Collection["Expr"]  # Callable args
 ParamDecl = "Declare"  # ProcFunc params (in statement)
 # HACK: Should use TypeAlias but not yet supported in Python 3.8
 Param = Union["TypedValue"]  # Callable params (in the frame)
@@ -433,7 +433,7 @@ class Builtin(PseudoValue):
     __slots__ = ('params', 'func')
     def __init__(
         self,
-        params: Iterable,
+        params: Collection[Param],
         func: function,
     ) -> None:
         self.params = params
@@ -465,7 +465,7 @@ class Callable(PseudoValue):
     def __init__(
         self,
         frame: "Frame",
-        params: Iterable[Param],
+        params: Collection[Param],
         stmts: Iterable["Stmt"],
     ) -> None:
         self.frame = frame
@@ -567,7 +567,7 @@ class Declare(Expr):
         self,
         name: NameKey,
         type: Type,
-        metadata: Mapping=None,
+        metadata: Mapping,
         *,
         token: "Token",
     ) -> None:
@@ -666,7 +666,7 @@ class GetIndex(Expr):
     __slots__ = ('array', 'index')
     def __init__(
         self,
-        array: NameExpr,
+        array: NameKeyExpr,
         index: IndexExpr,
     ) -> None:
         self.array = array
@@ -851,8 +851,8 @@ class TypeStmt(Stmt):
     __slots__ = ('name', 'exprs')
     def __init__(
         self,
-        name: NameExpr,
-        exprs: Iterable["Expr"],
+        name: Name,
+        exprs: Iterable["Declare"],
     ) -> None:
         self.name = name
         self.exprs = exprs
--- a/pseudocode/lang.py
+++ b/pseudocode/lang.py
@@ -829,7 +829,7 @@ class ProcFunc(Stmt):
     __slots__ = ('name', 'passby', 'params', 'stmts', 'returnType')
     def __init__(
         self,
-        name: GetExpr,
+        name: Name,
         passby: LiteralType['BYVALUE', 'BYREF'],
         params: Iterable[Declare],
         stmts: Iterable["Stmt"],
--- a/pseudocode/lang.py
+++ b/pseudocode/lang.py
@@ -13,6 +13,7 @@ NameKey = str  # Key for Object/Frame
 IndexKey = Tuple[int, ...]  # Key for Array
 IndexExpr = Tuple["Literal", ...]  # Array indexes
 IndexRange = Tuple["Literal", "Literal"]  # Array ranges (declared)
+Passby = LiteralType['BYREF', 'BYVALUE']
 Args = Iterable["Expr"]  # Callable args
 ParamDecl = "Declare"  # ProcFunc params (in statement)
 # HACK: Should use TypeAlias but not yet supported in Python 3.8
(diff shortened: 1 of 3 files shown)


ngjunsiang:
Our ExprStmt subclasses are still expecting an Expr, which wreaks havoc with the resolver and interpreter which expect more specific Expr subtypes. So let's give them their individual parameter types:
https://github.com/nyjc-computing/pseudo-9608/blob/8de9d2a4172f5a2bfd2d8a49a1847bf9eadf339b/pseudocode/lang.py#L727-L749

Code is much longer, but we'll shrinkwrap it later.

ngjunsiang:
Let's extend the stricter subtyping to other subclasses as well: [6e0c9b671c63c5784a9d115385c554b70841f886]

ngjunsiang:
# Resolver type issues

Now, on to the lengthy list of resolver issues that still remain.

ngjunsiang:
## verify() return types

```
pseudocode/resolver.py:361: error: Argument 1 to "expectTypeElseError" has incompatible type "None"; expected "str"
```
That's happening here:
https://github.com/nyjc-computing/pseudo-9608/blob/2f88622275ed162f2419652be591154e6943962a/pseudocode/resolver.py#L356-L362

First, our `verify()` helper is type-confused; sometimes it returns a Type and sometimes it returns `None`, depending on what kind of statement is parsed. We can help `mypy` out here by annotating return types more specifically with the `overload` decorator:
https://github.com/nyjc-computing/pseudo-9608/blob/2f88622275ed162f2419652be591154e6943962a/pseudocode/resolver.py#L436-L440

ngjunsiang:
## Type narrowing

```
pseudocode/resolver.py:361: error: Argument 1 to "expectTypeElseError" has incompatible type "None"; expected "str"
pseudocode/resolver.py:361: error: "Stmt" has no attribute "returnType"
pseudocode/resolver.py:361: error: "Stmt" has no attribute "name"
...
```
This is happening at:
https://github.com/nyjc-computing/pseudo-9608/blob/2f88622275ed162f2419652be591154e6943962a/pseudocode/resolver.py#L359-L362

We understand the logic here: if `stmt` is a Return Stmt, type-check it and raise an error if it doesn't work out. `mypy` doesn't understand this, because it is a static checker and doesn't run the code in `isReturn()`; it just checks for argument and return type.

Looks like we'll have to unwind some helpers, particularly `isReturn()`, `isProcedure()`, `isFunction()`. Not a big loss in return for static type checking, imo: [cb2a1dda3e92dddbfbd4bcab38f05653f22d1ee2]

ngjunsiang:
## Arrays and declarations

```
pseudocode/resolver.py:131: error: Missing positional arguments "ranges", "type" in call to "Array"
pseudocode/resolver.py:132: error: Value of type "Optional[Mapping[Any, Any]]" is not indexable
pseudocode/resolver.py:133: error: Value of type "Optional[Mapping[Any, Any]]" is not indexable
pseudocode/resolver.py:134: error: "Array" has no attribute "declare"
pseudocode/resolver.py:139: error: Item "None" of "Optional[Frame]" has no attribute "getType"
pseudocode/resolver.py:142: error: Item "None" of "Optional[Frame]" has no attribute "get"
```
https://github.com/nyjc-computing/pseudo-9608/blob/cb2a1dda3e92dddbfbd4bcab38f05653f22d1ee2/pseudocode/resolver.py#L117-L141

We changed Array so that it declares its indexes in the init, instead of presenting a `declare() method. So those changes need to be propagated to the resolver too.

We start by updating `verifyDeclareType()`:
https://github.com/nyjc-computing/pseudo-9608/blob/58cf993493e0bd6f709c004d7cca2765531d9ca5/pseudocode/resolver.py#L442-L447

This now declares attributes in an ObjectTemplate instead of an Object, with the help of `resolveDeclare()`. So that helper needs updating too:
https://github.com/nyjc-computing/pseudo-9608/blob/947204bbccf42b537045cdf3afa8dcf14c402d8b/pseudocode/resolver.py#L119-L153

We do our checks to ensure we are not declaring ARRAYs in a TYPE statement block since our interpreter can't handle that yet. And we add some `assert` checks to reassure `mypy` that we are not attempting to do anything funny like handle a `frame.outer` that is `None`.

ngjunsiang:
## Get*s

```
pseudocode/resolver.py:225: error: Argument 1 to "resolveGetName" has incompatible type "Union[Union[UnresolvedName, GetName, GetAttr, GetIndex], Call]"; expected "Frame"
pseudocode/resolver.py:225: error: Argument 2 to "resolveGetName" has incompatible type "Name"; expected "GetName"
pseudocode/resolver.py:228: error: Argument 2 to "declaredElseError" has incompatible type "str"; expected "Name"
pseudocode/resolver.py:232: error: "TypeSystem" has no attribute "clone"
pseudocode/resolver.py:254: error: "GetIndex" has no attribute "arrayExpr"
pseudocode/resolver.py:255: error: Incompatible types in assignment (expression has type "Union[float, str, PseudoValue, None]", variable has type "Array")
pseudocode/resolver.py:255: error: "GetIndex" has no attribute "frame"
pseudocode/resolver.py:262: error: Argument 1 to "getType" of "Object" has incompatible type "Name"; expected "str"
pseudocode/resolver.py:264: error: Missing return statement
pseudocode/resolver.py:268: error: Missing named argument "token" for "resolveAttr"
```

Now let's tackle the Get-related issues. One source of type conflict is that we are trying to handle name existence in an ObjectTemplate and in a TypeSystem with a single helper, `declaredElseError()`. Instead of making more helpers, let's remove this one first: [586c50edc368c56eddc318a1236a1639cfc86fea], [43c510f4f08b58353f789a8afba711278a90054c]

More Array type-checking to make `mypy` happy: [2ee8138129fc503a0c52a3695352b87618444885]

ngjunsiang:
## Of Names and Gets

```
pseudocode/resolver.py:320: error: Missing return statement
pseudocode/resolver.py:334: error: Module has no attribute "Get"
```

Our attempt to shoehorn GetName, GetAttr, GetIndex, and similar gets into a single type seems to be doing more harm than good. Let's throw them all into `resolve()` as separate cases and get rid of `resolveGet()`: [e47d794467935d00d7c5983ce242841e5a065fca]

Then we have a whole headache around ensuring that we catch every single UnresolvedName and turn it into a GetName. We have a helper for doing that:
https://github.com/nyjc-computing/pseudo-9608/blob/e47d794467935d00d7c5983ce242841e5a065fca/pseudocode/resolver.py#L46-L67

This will, given an Expr or Stmt and an attribute string, check if the attribute of the expr/Stmt is an UnresolvedName, and change it to a GetName if so. But we still have to do the heavy lifting of determining where our potential UnresolvedNames are, and passing them to `resolveName()`. Time to refactor `resolveName()` and add another helper:
https://github.com/nyjc-computing/pseudo-9608/blob/60dde6e3583fa3c0698fa9de9a2118daef77ffa6/pseudocode/resolver.py#L46-L67
https://github.com/nyjc-computing/pseudo-9608/blob/60dde6e3583fa3c0698fa9de9a2118daef77ffa6/pseudocode/resolver.py#L69-L79

Lots of checks and type-narrowing to placate `mypy` in `resolveName()`, not forgetting to return the GetName because `resolveExprs()` needs it:
https://github.com/nyjc-computing/pseudo-9608/blob/e47d794467935d00d7c5983ce242841e5a065fca/pseudocode/resolver.py#L338-L345

And then, we add a brute-force helper that simply passes every attribute in an Expr/Stmt through `resolveName()` so we ensure we don't miss any out. Then we use it wherever we can: [177a02449451dfaf8a1496b186a3040b181f621e]

There, no more anxiety over unchecked UnresolvedNames!

ngjunsiang:
We have whittled down hundreds of type errors down to 83! Let's keep pushing.

## Managing Iterables

```
pseudocode/resolver.py:374: error: Argument 1 to "len" has incompatible type "Iterable[Expr]"; expected "Sized"
pseudocode/resolver.py:375: error: Value of type "Iterable[Expr]" is not indexable
pseudocode/resolver.py:376: error: Unsupported target for indexed assignment ("Iterable[Expr]")
pseudocode/resolver.py:376: error: Value of type "Iterable[Expr]" is not indexable
pseudocode/resolver.py:377: error: Value of type "Iterable[Expr]" is not indexable
```
That happens here:
https://github.com/nyjc-computing/pseudo-9608/blob/177a02449451dfaf8a1496b186a3040b181f621e/pseudocode/resolver.py#L370-L377

And it happens because `expr` is of type `Iterable[lang.Expr]`. In python, iterables are objects that you can iterate through, e.g. in a `for` loop, but without necessarily knowing how many elements there are, so it doesn't support use in `len()`. For that, we need a Collection type ... or we could just do it a different way and stick with iterables: [ca7487549cf6be92f1c43d76867da98d0f56f01a]
https://github.com/nyjc-computing/pseudo-9608/blob/ca7487549cf6be92f1c43d76867da98d0f56f01a/pseudocode/resolver.py#L371-L390

This way, we don't have to worry about indexes, and instead of doing an in-place modification of `exprs` we return a modified iterable of them that get re-assigned: [04ae9429dc1c677dd46b665922432b7fdc767607]

It's also used in one place for now but we might need it later, e.g. in `resolve*Call()`: [77592c12170bf92b20268904ea237a128cdece63]

ngjunsiang:
Since we've landed on the Procedure/Function calls, let's clean up here as well: [25776cf7a84c6c624585bae6df2af251f8c40a0c]

And remove unused code: [354e19f013b5d0595682a2ed40f78f70a8b4d9c8]

And then we can fix type errors in `verifyProcedure()` and `verifyFunction()`:
```
pseudocode/resolver.py:436: error: Argument 2 to "resolveDeclare" has incompatible type "Tuple[int, Declare]"; expected "Declare"
pseudocode/resolver.py:437: error: "Tuple[int, Declare]" has no attribute "name"
pseudocode/resolver.py:445: error: Argument 1 to "declare" of "Object" has incompatible type "Union[UnresolvedName, GetName, GetAttr, GetIndex]"; expected "str"
pseudocode/resolver.py:446: error: Argument 1 to "setValue" of "Object" has incompatible type "Union[UnresolvedName, GetName, GetAttr, GetIndex]"; expected "str"
pseudocode/resolver.py:449: error: "Callable[[object], TypeGuard[Callable[..., object]]]" has no attribute "stmts"
pseudocode/resolver.py:462: error: Argument 1 to "declare" of "Object" has incompatible type "Union[UnresolvedName, GetName, GetAttr, GetIndex]"; expected "str"
pseudocode/resolver.py:463: error: Argument 1 to "setValue" of "Object" has incompatible type "Union[UnresolvedName, GetName, GetAttr, GetIndex]"; expected "str"
```

ngjunsiang:
## Functions and Procedures

Let's do a round of type-fixing and adding missing arguments: [afd574aa91bd1355175302b45fbfa5ab387af581]

Looks like `verifyProcedure()` and `verifyFunction()` are basically fine, apart from being confused about what to do with ProcFunc.name. Fine, since Procedures and Functions are declared in statements and there is no reason to ever treat them as a GetName, let's treat them as Names instead: [60b5fd9be40b3214a04fa251791455ad5ad943eb]

That means to use them as a NameKey we need to `str()` them: [a01bd38e333587880b26471106607e0e5695c56b]

`transformDeclare()` needs some fixing and readability improvements too: [08fabc4b8361a2a400d25a0a7adabde319db37e3]

ngjunsiang:
# Parser (redux)

Let's wrap up with a couple of type fixes we left out in the parser.

First, allowing Array to accept NameExprs for the `array`; we don't allow Arrays coming from a GetAttr but we've used `asserts and other guards to prevent that happening: [5d48f70f3e433e13514f5c61f7d7f0e407921c0b]

Next, our little issue with metadata: `mypy` really does not like arbitrary dicts with heterogeneous types. So let's make a TypedDict:
https://github.com/nyjc-computing/pseudo-9608/blob/4c8f26956ba484dc042f5d72ce450791b6053d23/pseudocode/lang.py#L31-L34

And use it:
https://github.com/nyjc-computing/pseudo-9608/blob/4c8f26956ba484dc042f5d72ce450791b6053d23/pseudocode/parser.py#L249-L259

Plus miscellaneous fixes I had to make: [4c8f26956ba484dc042f5d72ce450791b6053d23]

Great, we cleaned up all `mypy`'s reported type errors in the parser too 🥳

Final cleaning: [173a5d728395af1f13179a6fcc26252930103824]

Turns out we didn't need our PseudoMap protocol after all. Good thing I didn't pour that much time into it!

PR #70, opened 2022-05-31
24f Improvements: Interpreter

ngjunsiang:
Finally we get to the interpreter, where we have ... **\*checks notes\*** 39 out of 51 errors to fix (the remaining coming from `__init__.py`).

So let's get to it. But first, some cleaning up for formatting: [dfaaff7014e41eb94bbd66ac93de39bd433aee9d]

### DIFF
--- a/pseudocode/interpreter.py
+++ b/pseudocode/interpreter.py
@@ -12,15 +12,14 @@ def expectTypeElseError(
     exprmode: str,
     *expected: str,
     errmsg: str="Expected",
-    token: lang.Token=None,
+    token: lang.Token,
 ) -> None:
     if not exprmode in expected:
-        if not token: token = exprmode
-        raise builtin.RuntimeError(f"{errmsg} {expected}", token)
+        raise builtin.RuntimeError(f"{errmsg} {expected}", token=token)
 
 def declaredElseError(
     frame: lang.Frame,
-    name: lang.Varname,
+    name: lang.NameKey,
     errmsg: str="Undeclared",
     token: lang.Token=None,
 ) -> None:
@@ -29,7 +28,7 @@ def declaredElseError(
 
 def undeclaredElseError(
     frame: lang.Frame,
-    name: lang.Varname,
+    name: lang.NameKey,
     errmsg="Already declared",
     token: lang.Token=None,
 ) -> None:
--- a/pseudocode/interpreter.py
+++ b/pseudocode/interpreter.py
@@ -120,12 +120,9 @@ def evalCall(frame: lang.Frame, expr: lang.Call, **kwargs) -> Optional[lang.Valu
         for arg, slot in zip(expr.args, callable.params):
             argval = evaluate(frame, arg)
             slot.value = argval
-        for stmt in callable.stmts:
-            returnval = execute(callable.frame, stmt, **kwargs)
-            if returnval is not None:
-                return returnval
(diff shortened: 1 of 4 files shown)


ngjunsiang:
Starting with some type fixes; Python provides a `NoReturn` return type for functions that should not return anything, so let's use it: [ee705797395b9303301cb090594585349df21fc5]

Next, to make our return types work, 1execute()` should not return anything. That means any Return Stmts (which would be evaluated in evalCall()` should not be handled indirectly via `execute()`, but directly via `execCall()`: [ce0c136fbf66b38632b8fcf0fe80612223f1d06b]

Since this will always be coming from the `stmts` block of a Callable, we can delegate it to `executeStmts()`:
https://github.com/nyjc-computing/pseudo-9608/blob/ce0c136fbf66b38632b8fcf0fe80612223f1d06b/pseudocode/interpreter.py#L158-L167

We detect Return statements manually, and return them directly. This lets us shorten the code in `evalCall()`, where we can just return directly from `executeStmts()`:
https://github.com/nyjc-computing/pseudo-9608/blob/ce0c136fbf66b38632b8fcf0fe80612223f1d06b/pseudocode/interpreter.py#L109-L123

ngjunsiang:
## Minor fixes

Let's begin with some minor fixes: fixing incorrect type names, optional arguments, etc: [9b69cf7151b9a6ecc8176dc1098569e0974d118e]

# No more Get

Our first problem is that there is no more Get Expr; it has been broken up into GetName, GetIndex, and GetAttr. So let's start with `evaluate()`: [87af983c9d5d4dc9cc33aafbac24f7c1fffe1888]

We pretty much have to rewrite `evalGet()`:
https://github.com/nyjc-computing/pseudo-9608/blob/c5fc048ec61e9d58911c96201b0e442ee11577ee/pseudocode/interpreter.py#L95-L106

`evalCall()` is okay, but `mypy` still wants to see our `assert`s to be reassured of its types: [ddc915f668d45caaaeee6600e98f9bf5dacfb6be], [b63d678fc7efe629a929ac272b583a7e82bf2f10]

`evalAssign()` and `execInput()` are also affected by the Get evolution, so let's look at them in more detail.
https://github.com/nyjc-computing/pseudo-9608/blob/302d290a80360d6dd509eda17b0099a635fc004b/pseudocode/interpreter.py#L130-L147

We have to differentiate between Frames, Arrays, and Objects, and treat them differently since they use different keys; not much to it. And we raise an error at the end to placate `mypy` so it doesn't think we are implicitly returning `None`.

https://github.com/nyjc-computing/pseudo-9608/blob/302d290a80360d6dd509eda17b0099a635fc004b/pseudocode/interpreter.py#L201-L220

Same story for `execInput()`; it takes user input and assigns it to a given GetExpr. So the code is going to look quite similar to `evalAssign()`; we'll refactor this with a helper in future.

ngjunsiang:
# Indexes

Our last fix is for `evalIndex()`; there's nothing wrong with the code functionally. We have type-checked the expressions in the resolver already, but `mypy` still needs reassurance that the indexes are `int`s. So let's rewrite that:
https://github.com/nyjc-computing/pseudo-9608/blob/e922d476f78209d9743a12bb4741a1ad3d58ec15/pseudocode/interpreter.py#L72-L81

ngjunsiang:
We are almost done with propagating our changes. Next chapter: Pseudo itself!

PR #73, opened 2022-06-02
25b Automation: setup

ngjunsiang:
Nice, nice, things are coming along really fine. Our user, presumably a Python programmer, can install our `pseudo-9608` package, then import the `pseudocode` module, and pass a string to `pseudocode.run()` or a filename to `pseudocode.runFile()`  to execute pseudocode.

This is a hassle though, and certainly not feasible for a user who does not know Python. It would be more ideal if an administrator could install the `pseudo-9608` package and have it provide a `pseudo` executable; running `pseudo mycode.pseudo` would then interpret the pseudocode in `mycode.pseudo`. This is how Python and other interpreters run, too.

Let's make it happen for `pseudo-8608`.


ngjunsiang:
The poetry documentation has a [scripts](https://python-poetry.org/docs/pyproject/#scripts) section that describes how to specify scripts to be installed when installing the package:

> ```
> [tool.poetry.scripts]
> poetry = 'poetry.console:run'
> ```
> Here, we will have the `poetry` script installed which will execute `console.run` in the poetry package.

Looks like we need a `run()` function as an entry point for this script; the `run()` function will be called when the script is invoked. We already have our code in [`pseudo`](https://github.com/nyjc-computing/pseudo-9608/pull/73/commits/5d9e2c9eae8d8bc2f8e5dc06c83d5daf80414f31#diff-e0b714bf90348b25b918e0a8998f5b7451ae5d276be20d46c30d4f4f72526470), so let's refactor it into a function: [5d9e2c9eae8d8bc2f8e5dc06c83d5daf80414f31]

ngjunsiang:
Now we'll have to include `pseudo` inside our package: [8bc9ab8fb8fb32e38aafb3f4b344d086d4f0ed77]

The Github Actions workflow we created in the previous chapter still seems to keep running even though I've marked the pull request as draft, so let's add a filter to stop that happening: [c78170ee707b696df2d5b27262fdb36c8a7d2e44]

And then we follow the Poetry documentation to specify our own `pseudo` script: [0862ca1f2f03dce14c2194ac01503cdc6367f8a4]
https://github.com/nyjc-computing/pseudo-9608/blob/0862ca1f2f03dce14c2194ac01503cdc6367f8a4/pyproject.toml#L20-L21

I also took the chance to move `mypy` to dev dependencies, since a typical user is not going to need it to use the package; it is only for people who wish to work on `pseudo-9608`!
https://github.com/nyjc-computing/pseudo-9608/blob/0862ca1f2f03dce14c2194ac01503cdc6367f8a4/pyproject.toml#L14-L18

ngjunsiang:
Okay, this didn't work in testing. I suspect Python is not detecting `pseudo` as a module in the `pseudocode` package, possibly because of that first line:
https://github.com/nyjc-computing/pseudo-9608/blob/5bdc8707c7f1721fb0985df1a3e8f59d1762154d/pseudocode/pseudo#L1

Also, `pseudo` attempts to import the `pseudocode` package, leading to a circular import:
https://github.com/nyjc-computing/pseudo-9608/blob/5bdc8707c7f1721fb0985df1a3e8f59d1762154d/pseudocode/pseudo#L4-L5

Let's rethink this. The `pseudo` module itself doesn't really do anything besides handling arguments; we can do that within `pseudocode/__init__.py` too. Let's merge `pseudo` into `__init__.py`: [4841833072f0cd74696dca2b19477dc2f566bcf5]

We also update the entry point in `pyproject.toml` to `'pseudocode:main'`, bump the package version, and republish it.

ngjunsiang:
Running pseudo in the shell:

```
$ pseudo
Johnson
Leroy
6
1
2
3
4
5
6
7
8
9
10
```

Great, it works 😎

But if `main.pseudo` is missing:
```
$ pseudo
  File "venv/bin/pseudo", line 8, in <module>
    sys.exit(main())
  File "venv/lib/python3.8/site-packages/pseudocode/__init__.py", line 120, in main
    result = pseudo.runFile(srcfile)
  File "venv/lib/python3.8/site-packages/pseudocode/__init__.py", line 62, in runFile
    with open(srcfile, 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: 'main.pseudo'
```

That's not nice; we don't want Python errors leaking through. But that's a problem for a different chapter.

ngjunsiang:
Looks like poetry's bundled publishing tools are sufficient to help us set up the script for the entry point. We have automated testing for each non-draft pull request, and poetry itself helps us with the publishing and installation.

PR #74, opened 2022-06-02
26a Releasing: Usage

ngjunsiang:
What's left? A bunch of features not yet implemented, and an official release announcement.

Before we get to that point, we want to get pseudo into useable state: a user who installs a package and runs `pseudo` should at least have it working as expected.

This chapter is all about getting to that point.

# A build script

To help us out with tedious repetitive tasks, I upgraded the test script to a build script: [35ad0136da01bd94ece8e6da6f43f520b8fec657]
https://github.com/nyjc-computing/pseudo-9608/blob/9ef71ebc9631207072efaf48adfc1f7527beb0c4/build#L34-L79

This is written in a scripting language, bash, and not Python. I won't be explaining this part in detail since it's not the focus here.


ngjunsiang:
# Usage help

Every commandline program provides some kind of help. Running `python -h` displays the following:
```
python -h
usage: /home/runner/pseudo-9608/venv/bin/python [option] ... [-c cmd | -m mod | file | -] [arg] ...
Options and arguments (and corresponding environment variables):
-b     : issue warnings about str(bytes_instance), str(bytearray_instance)
         and comparing bytes/bytearray with str. (-bb: issue errors)
-B     : don't write .pyc files on import; also PYTHONDONTWRITEBYTECODE=x
-c cmd : program passed in as string (terminates option list)
-d     : debug output from parser; also PYTHONDEBUG=x
-E     : ignore PYTHON* environment variables (such as PYTHONPATH)
-h     : print this help message and exit (also --help)
-i     : inspect interactively after running script; forces a prompt even
         if stdin does not appear to be a terminal; also PYTHONINSPECT=x
[... more lines truncated]
```

`pseudo` is not so advanced, but nonetheless should still display some kind of help for new users.

ngjunsiang:
Let's start by adding the message content: [dfdd2bec946c03f93a89d939479e295a2ee5b644]
https://github.com/nyjc-computing/pseudo-9608/blob/dfdd2bec946c03f93a89d939479e295a2ee5b644/pseudocode/__init__.py#L26-L31

We can `print(HELP)` when we want to display this help message.

Next, handling the `-h` flag.

ngjunsiang:
And then handling the `-h` flag: [96daf1164684571c015ae643d10cabacb901263d]
https://github.com/nyjc-computing/pseudo-9608/blob/96daf1164684571c015ae643d10cabacb901263d/pseudocode/__init__.py#L129-L131

ngjunsiang:
# Testing

Testing the use of the main script is different from testing python code. Instead of running the code, or running unittests or `mypy`, we have to run this in the shell. The easiest way I found to do this is to build the package and then install it, with poetry:

```
$ poetry build
[...]
$ poetry install
[...]
```

This makes the `pseudo` script available for running:
```
$ pseudo
No argument provided.
Try `pseudo -h' for more information.
$ pseudo -h
usage: pseudo [option] ... file
Options and arguments:
-h     : print this help message and exit (also --help)
file   : program read from script file
```

Looks like it worked great!

ngjunsiang:
# Unrecognized options

Okay, what about unrecognized options? These are arguments that start with `-` but which we don't handle ... yet: [bab627bb6aa8a59238a6a3e72b99a86e32d3135f]
https://github.com/nyjc-computing/pseudo-9608/blob/bab627bb6aa8a59238a6a3e72b99a86e32d3135f/pseudocode/__init__.py#L130-L133

```
$ pseudo -v
Unknown option: -v
Try `pseudo -h' for more information.
```

ngjunsiang:
# Invalid file

What if a filename is provided, but can't be read for some reason?

First let's check that it exists:
https://github.com/nyjc-computing/pseudo-9608/blob/770bde0b257580010686e6927a169ac1a0f5e0d3/pseudocode/__init__.py#L137-L139

The easiest way to ensure it can be read is to ... try to read it:
https://github.com/nyjc-computing/pseudo-9608/blob/ef94cebd80717d3102fb1d7c876b2cc440f4cf7d/pseudocode/__init__.py#L140-L146

Some Python is leaking through, since we don't catch specific types of errors and print custom messages for them. We just dump Python's error message. This will do for now; pseudo-ish error messages are a feature for the future.

ngjunsiang:
We now have something closer to what a user would expect for a commandline program; let's bump the version to 0.4.0! [9079ed76bb3abd9838725dfe61570c4f1791a35c]

And fix a bug in the build script that wasn't substituting the version correctly: [70ff246ca1fdd58ebcaef58ba4724b73bf1ec8b5]

PR #76, opened 2022-06-02
26b Release: REPL

ngjunsiang:
The next thing any programmer expects from a programming language is a REPL (read-eval-print loop). That is, they expect, when a program is launched without any arguments, to see an interactive prompt that lets them type statements, one line at a time. This is what the python REPL looks like:
```
$ python
Python 3.8.12 (default, Aug 30 2021, 16:42:10) 
[GCC 10.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> print('Hello World!')
Hello World!
```

Our users will be expecting pseudo to implement a REPL as well, so let's respond to that expectation. For now pseudo does not have a REPL, but by the end of this chapter it will!
https://github.com/nyjc-computing/pseudo-9608/blob/1fab24d9e7ff6a9b64d6b55aa7c546423083c095/pseudocode/__init__.py#L119-L122

### DIFF
--- a/pseudocode/__init__.py
+++ b/pseudocode/__init__.py
@@ -116,11 +116,12 @@ class Pseudo:
 # https://gist.github.com/bojanrajkovic/831993
 
 def main():
-    # Argument handling
+    # REPL
     if len(sys.argv) == 1:
-        print("No argument provided.")  # Unhandled error
-        print("Try `pseudo -h' for more information.")
-        sys.exit(64)  # command line usage error
+        print("REPL unsupported.")  # Unhandled error
+        sys.exit(0)
+
+    # Argument handling
     if len(sys.argv) > 1:
         if sys.argv[1] == '-h':
             print(HELP)
--- a/pseudocode/__init__.py
+++ b/pseudocode/__init__.py
@@ -36,7 +36,7 @@ def printException() -> None:
     info, error = traceback.format_exception(etype, value, tb)[-2:]
     print(f'Exception in:\n{info}\n{error}')
 
-def error(lines: Iterable[str], err: builtin.PseudoError) -> None:
+def report(lines: Iterable[str], err: builtin.PseudoError) -> None:
     errType = type(err).__name__ + ':'
     if err.line:
         lineinfo = f"[Line {err.line}]"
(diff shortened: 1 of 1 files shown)


ngjunsiang:
# REPL vs script

The key differences between REPL mode and script mode are:
1. in REPL mode, the user is continually prompted for more code, while script mode exits after running the script
2. The global frame persists after each statement in REPL mode, but resets between scripts
3. Any errors that occur in a script causes execution to halt, while errors in  REPL mode are simply reported without program exit

ngjunsiang:
# REPL mode

Let's start with the simplest possible REPL: a `while` loop that reads user input with `input()`, evaluates the input with `Pseudo.run()`, and prints the result with ... well, `print()`: [55c907831488d37d3347f589a5813e9b784f58d2]
https://github.com/nyjc-computing/pseudo-9608/blob/55c907831488d37d3347f589a5813e9b784f58d2/pseudocode/__init__.py#L118-L128

```
$ pseudo
Pseudo 0.4.0
### OUTPUT "Hello World!"                               
Hello World!
### 
```

Simpler than we thought! This is still lacking all the complexity required to handle other tasks, such as arrow key movement, command history, and multi-line statements. But it's a start!

ngjunsiang:
# Preserving global frame

The global frame is re-instantiated with every call to `Pseudo.run()`:
https://github.com/nyjc-computing/pseudo-9608/blob/55c907831488d37d3347f589a5813e9b784f58d2/pseudocode/__init__.py#L70-L80

This means that in our REPL, each line is running with a completely new frame. We wouldn't even be able to declare a variable and then assign to it.

The simplest way to keep the frame around after `run()` is to instantiate the frame once, when `Pseudo` is instantiated, and then keep using it. If we want a new frame, we just make a new instance of `Pseudo`.

Let's make those changes:
https://github.com/nyjc-computing/pseudo-9608/blob/7c8fa649bb0fc9b5fe1fe858ff70ee7d9812ae3b/pseudocode/__init__.py#L51-L61

```
$ pseudo
Pseudo 0.4.0
### DECLARE Variable : STRING
### Variable <- "Hello World!"
### OUTPUT Variable
Hello World!
### 
```

And now we can `run()` single lines with a persistent frame.

ngjunsiang:
# Report without exit

We actually already accomplished this earlier in **REPL Mode**; by simply reporting errors without calling `sys.exit()`, our REPL continues to run instead of exiting.

ngjunsiang:
So there are a bunch of issues to be addressed, but not in this chapter:
- #77 REPL Mode should enable cursor navigation (with arrow keys)
- #78 Command history for REPL Mode
- #79 Multi-line statement support for REPL Mode

PR #80, opened 2022-06-03
26c Release: Logging

ngjunsiang:
We are almost ready to go! But once pseudo is released, someone is sure to run into a bug at some point, and they wouldn't know how to report it ... we'll need logs that they can submit. Which means implementing logging.

# Logging unexpected errors

We'll start simple, with an example coming from https://realpython.com/python-logging/:
https://github.com/nyjc-computing/pseudo-9608/blob/ec6d13e32807cff4ee2d68ffadcf678c9f7f6247/pseudocode/__init__.py#L6-L11

And we use `logging.exception()` to capture error information to the file: [ec6d13e32807cff4ee2d68ffadcf678c9f7f6247], [5ee99590e5a26350a0cf79e8df74ffe952170ea5]

### DIFF
--- a/pseudocode/__init__.py
+++ b/pseudocode/__init__.py
@@ -28,7 +28,7 @@ class Result(TypedDict):
 
 
 
-__version__ = '0.4.0'
+__version__ = '0.4.1'
 VERSION = f"Pseudo {__version__}"
 HELP = """
 usage: pseudo [option] ... file
--- a/pseudocode/__init__.py
+++ b/pseudocode/__init__.py
@@ -1,7 +1,7 @@
 from typing import Optional
 from typing import Iterable, List, MutableMapping
 from typing import TypedDict, Callable as function
-import os, sys, traceback
+import os, sys
 
 import logging
 logging.basicConfig(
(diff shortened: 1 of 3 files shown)


ngjunsiang:
I'm going to do a quick test by adding an `assert False` to trigger an `AssertionError`: 

When running pseudo I get this in `pseudo.log`:
```
root - ERROR - Unexpected error has occurred
  File "/home/runner/pseudo-9608/pseudocode/__init__.py", line 113, in run
    interpreter.interpret()
  File "/home/runner/pseudo-9608/pseudocode/interpreter.py", line 64, in interpret
    assert False
AssertionError
```

Glad it worked :sweat_smile:

Let's clean up: [66fd5ce81a9361f4cf3efc6e559a25a9bbf1beba]

ngjunsiang:
# Informing users about unexpected errors

It's not nice to fail silently and keep the user guessing as to what happened. At the same time, we don't want to dump the gory details on them and expect them to clean up. Let's include some bug-reporting instructions:[9dac5cb9f3e5c1961a44452ea36c1031300a1955]
https://github.com/nyjc-computing/pseudo-9608/blob/9dac5cb9f3e5c1961a44452ea36c1031300a1955/pseudocode/__init__.py#L42-L51

ngjunsiang:
There is much more logging to be done, but let's not add more code in anticipation. We'll add that when the need arises.

We are almost ready to launch! Just one more thing ...

PR #81, opened 2022-06-03
26d Release: Readme

ngjunsiang:
Our users are going to have lots of questions, and if we don't want to be overwhelmed with questions, we should provide reading material that explains common questions in an accessible way. These are the readmes which we will be adding in this chapter.

For a start, the chapters that I write as pull requests are linked in the main README. They have gotten lengthy enough that I think it is time to split them off to a separate file, so as to keep `README.md` readable: [95adfd73d8d91c0bd9a73b4ab011a9ee8d7d615c]

And then some updates to `README.md` to explain its use: [87958e0451c82b39aa255432e0ba88aeacd85333]


ngjunsiang:
# Reference Guide

9608 pseudocode syntax is used for academic purposes, so its use is likely to be explained by a Computing teacher. As such, I will dispense with lengthier tutorials and examples explaining common patterns of programming in this language.

However, we might have some users who are not familiar with this style of pseudocode at all, so a reference guide is still handy: [44a12480ced6a63ee3b02fd34caaf3f3b44c7858]

ngjunsiang:
# Roadmap

Some users are bound to ask if X feature will be implemented, or why Y feature won't be implemented. Let's include a roadmap for them: [808d7973c04ea209c0d76ba246fa42c9d3d498be]

ngjunsiang:
With that, we are about 90% done :) for a minimum viable interpreter, anyway. Time to release an alpha!

PR #82, opened 2022-06-07
27 Documentation: Docstrings

ngjunsiang:
Now that we have a release, it’s time to prep the codebase for potential contributions. Besides making things easier for contributors, it also makes it easier for us/me to understand the code if I return to it in future.

First up: documentation.


ngjunsiang:
Even before we write the kind of lengthy detailed documentation that developers usually expect, I would start by writing docstrings first. That's the first instance of documentation that anyone trying to contribute code will first encounter.

So let's add docstrings for `__init__.py`, `builtin.py`, and `lang.py`: [7dbc483b39a024a538b7f4d316d60b2bd2596a9c]. [57fe719328fb8f7e7336d1b5dd35f675ab5c7561]

ngjunsiang:
That covers the main classes and functions/methods. Let's add docstrings for the modules too: [62b1660d9355d5c4b3b64c59cfe370898ece1d69]

I'm not documenting them exhaustively here since it's unlikely anyone will do `import pseudocode` and proceed to use the classes there, but we'll see.

ngjunsiang:
[PEP257](https://peps.python.org/pep-0257/) has some ideas about how docstrings should be formatted. We'll try to follow it: [ebdee49ec71188bda9ac7b552c1b7aec76bcbb31]

ngjunsiang:
And then docstrings for the rest of the modules: [e514b657908a5a9b3c71ca4ef12db0db1e751a6f]

This is probably the driest chapter so far, but also the one most likely to help us in future.

PR #83, opened 2022-06-07
28 Shrinkwrap: dataclasses

ngjunsiang:
This chapter is my favourite! Up to this point I have tried to "show the code" as much as possible, so it's easy to understand what is going on. That means not relying on advanced Python functions or classes with inscrutable implementation (yes, we can always read the source code, that's not the point though ...)

This results in much more verbose code than we can get away with. Now we have most of the baseline code in place; interested readers can revisit earlier chapters for that. I'm going to "shrink-wrap" the code by replacing them with Python's conveniences where available, provided their functionality is easy to understand.

First up: dataclasses.

### DIFF
--- a/pseudocode/builtin.py
+++ b/pseudocode/builtin.py
@@ -10,10 +10,10 @@ class PseudoError(Exception):
         super().__init__(msg)
         self.token = token
         self.line = line
-        self.col = None
+        self.column = None
         if token:
             self.line = token.line
-            self.col = token.col
+            self.column = token.column
         else:
             self.line = None
 
--- a/pseudocode/interpreter.py
+++ b/pseudocode/interpreter.py
@@ -7,7 +7,7 @@ execute(frame: Frame, statements: list) -> None
 from typing import Optional, Union
 from typing import Iterable, Callable as function
 from typing import overload
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from . import builtin, lang, system
 
@@ -58,10 +58,9 @@ def undeclaredElseError(
 @dataclass
 class Interpreter:
     """Interprets a list of statements with a given frame."""
-    __slots__ = ('frame', 'statements', 'outputHandler')
-    outputHandler: function = print
     frame: lang.Frame
     statements: Iterable[lang.Stmt]
+    outputHandler: function = field(default=print, init=False)
 
     def registerOutputHandler(self, handler: function) -> None:
         """Register handler as the function to use to handle any output
--- a/pseudocode/interpreter.py
+++ b/pseudocode/interpreter.py
@@ -184,7 +184,7 @@ def evalAssign(frame: lang.Frame, expr: lang.Assign) -> lang.Value:
         obj.setValue(name, value)
     else:
         raise builtin.RuntimeError(
-            "Invalid Input assignee", token=expr.assignee.token()
+            "Invalid Input assignee", token=expr.assignee.token
         )
     return value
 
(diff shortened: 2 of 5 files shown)


ngjunsiang:
# Python's Dataclass

So many of our classes actually don't *do* much; they have few/no methods attached, and primarily serve as **containers** for data: Token, Expr and its subclasses, Stmt and its subclasses, ...

For these classes we have a whole bunch of code that looks like:
```
class Token:
    def __init__(
        self,
        line: int,
        column: int,
        type: Type,
        word,
        value,
    ) -> None:
        self.line = line
        self.column = column
        self.type = type
        self.word = word
        self.value = value
```

For such classes, Python provides the [dataclass](https://docs.python.org/3.8/library/dataclasses.html#dataclasses.dataclass) decorator to reduce the above to:
https://github.com/nyjc-computing/pseudo-9608/blob/b00fb8138b84ac7e457fea9eaba20489943c54d8/pseudocode/lang.py#L63-L78

ngjunsiang:
Applying the `@dataclass` decorator to our TypedValues, templates, and PseudoValues: [2734a1b46c83054d34af141a7f333edde77b090e]

Expr is going to be a bit trickier: its `__init__()` takes a `token` named argument, but assigns it to the `_token` attribute, a pattern not supported by `dataclass`. We could simply not use dataclasses here ... but I think we can work around that by turning `token` into a [property](https://realpython.com/python-property/):
https://github.com/nyjc-computing/pseudo-9608/blob/37cebf7152ab59f218b8b84743aa1abb0e7fa52a/pseudocode/lang.py#L509-L525

This means that instead of implementing a `token()` method, all Expr subclasses need either a `token` property or attribute. And we can do that easily: [37cebf7152ab59f218b8b84743aa1abb0e7fa52a]

This will involve sweeping but minor changes throughout our code, turning all `token()` calls into `token` attribute gets from the parser onwards: [a3a006e251853aa00e7d1b83438f79d13411426c], [a73e706e9d12eacee84f77511decf2ae3d26738e]

ngjunsiang:
Dataclasses helped us shrink the code in `lang.py` by almost 20%! It is of limited use in the other modules though, where the classes primarily have a functional purpose rather than serving as wrappers around data.

ngjunsiang:
`outputHandler` in the Interpreter class is not required for init though; it defaults to the built-in `print` function. So we use `dataclasses.fields` to indicate that: [6400a899e8b7bf1af2c4a3d35d3c6be3269d4ce2]
https://github.com/nyjc-computing/pseudo-9608/blob/6400a899e8b7bf1af2c4a3d35d3c6be3269d4ce2/pseudocode/interpreter.py#L58-L77

ngjunsiang:
And we're done applying dataclasses. Checking that everything is still working:

```
$ python -m unittest
.........
Ran 9 tests in 0.157s

OK
```

ngjunsiang:
So we'll add some class attributes in Loop (which will get overridden by While and Repeat anyway) to make it happy: [685e02194d060726c24839478c925ef149d618f0]

This happens because `verifyLoop()` in the resolver applies the same checks to all Loop subclasses, but isn't aware that they all share the same attributes.

At the same time we missed out a `col` -> `column` rename, thank you for catching that `mypy`.

We'll ignore the incompatible assignment in the parser, since there isn't any feasible way for us to coerce the word from the passby token into a literal value other than more `assert` checks which I don't think are helpful here: [866becff83da115966cf1ed06eb02c8038472412]

PR #84, opened 2022-06-08
28b Shrinkwrap: Single dispatch

ngjunsiang:
Our resolver and interpreter use a technique known as **single dispatch**. This technique involves taking in an argument, and handling it differently depending on its type.

The current code for the resolver looks like this:
https://github.com/nyjc-computing/pseudo-9608/blob/922d15f963cbfeae36c0bfe6ed8c0d26a4038ef2/pseudocode/resolver.py#L382-L407

### DIFF
--- a/pseudocode/interpreter.py
+++ b/pseudocode/interpreter.py
@@ -466,43 +466,67 @@ def execReturn(
 
 
 
+@singledispatch
 def execute(
     stmt: lang.Stmt,
     frame: lang.Frame,
     **kwargs,
 ) -> None:
     """Dispatcher for statement executors."""
-    if isinstance(stmt, lang.Output):
-        execOutput(stmt, frame, **kwargs)
-    elif isinstance(stmt, lang.Input):
-        execInput(stmt, frame, **kwargs)
-    elif isinstance(stmt, lang.Case):
-        execCase(stmt, frame, **kwargs)
-    elif isinstance(stmt, lang.If):
-        execIf(stmt, frame, **kwargs)
-    elif isinstance(stmt, lang.While):
-        execWhile(stmt, frame, **kwargs)
-    elif isinstance(stmt, lang.Repeat):
-        execRepeat(stmt, frame, **kwargs)
-    elif (
-        isinstance(stmt, lang.OpenFile)
-        or isinstance(stmt, lang.ReadFile)
-        or isinstance(stmt, lang.WriteFile)
-        or isinstance(stmt, lang.CloseFile)
-    ):
-        execFile(stmt, frame, **kwargs)
-    elif isinstance(stmt, lang.CallStmt):
-        execCall(stmt, frame, **kwargs)
-    elif isinstance(stmt, lang.AssignStmt):
-        execAssign(stmt, frame, **kwargs)
-    elif isinstance(stmt, lang.Return):
-        execReturn(stmt, frame, **kwargs)
-    elif (
-        isinstance(stmt, lang.DeclareStmt)
-        or isinstance(stmt, lang.TypeStmt)
-        or isinstance(stmt, lang.ProcedureStmt)
-        or isinstance(stmt, lang.FunctionStmt)
-    ):
-        pass
-    else:
-        raise ValueError(f"Invalid Stmt {stmt}")
+    raise TypeError(f"Invalid Stmt {stmt}")
+
+@execute.register
+def _(stmt: lang.Output, frame: lang.Frame, **kwargs) -> None:
+    execOutput(stmt, frame, **kwargs)
+
+@execute.register
+def _(stmt: lang.Input, frame: lang.Frame, **kwargs) -> None:
+    execInput(stmt, frame, **kwargs)
+
+@execute.register
+def _(stmt: lang.Case, frame: lang.Frame, **kwargs) -> None:
+    execCase(stmt, frame, **kwargs)
+
+@execute.register
+def _(stmt: lang.If, frame: lang.Frame, **kwargs) -> None:
+    execIf(stmt, frame, **kwargs)
+    
+@execute.register
+def _(stmt: lang.While, frame: lang.Frame, **kwargs) -> None:
+    execWhile(stmt, frame, **kwargs)
+    
+@execute.register
+def _(stmt: lang.Repeat, frame: lang.Frame, **kwargs) -> None:
+    execRepeat(stmt, frame, **kwargs)
+    
+@execute.register
+def _(stmt: lang.FileStmt, frame: lang.Frame, **kwargs) -> None:
+    execFile(stmt, frame, **kwargs)
+    
+@execute.register
+def _(stmt: lang.CallStmt, frame: lang.Frame, **kwargs) -> None:
+    execCall(stmt, frame, **kwargs)
+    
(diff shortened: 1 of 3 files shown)


ngjunsiang:
Notice what it is doing: checking the type of the Expr, and then delegating the job to an appropriate resolver.

The code is fine as it is, but from a performance standpoint[^1] this is suboptimal; each call to `resolve()` involves up to 9 type-checks before an appropriate resolver is found.

[^1]: It feels strange to be optimising for performance in an interpreted language like Python; clearly the most optimal solution is to use a compiled binary. Nonetheless, there are still real performance gains to be made especially in a function that gets called so often.

ngjunsiang:
# The single dispatch pattern

Why don't we do the type-check once, pick an appropriate resolver, and then use it, without needing a sequence of checks? Something like this: [f7aece75cd35c5bee987ba4c63ab3e25c6425f26], [d00ba7bb54a6ffbac330f498bc64f79569990872]
https://github.com/nyjc-computing/pseudo-9608/blob/d00ba7bb54a6ffbac330f498bc64f79569990872/pseudocode/resolver.py#L382-L402

Python's `dict` type lets us use any hashable object as a key, including classes. Thus, the `resolver` dict lets us map Expr classes to their resolvers, and we use `resolver` in `resolve()` to do just this.

ngjunsiang:
# functools.singledispatch

We can use Python's [`singledispatch`](https://docs.python.org/3/library/functools.html#functools.singledispatch) decorator from `functools` to enjoy the typing benefits: https://medium.com/@sunilrana123/python-trick-functools-singledispatch-4cdf71843cd2

But `singledispatch` only dispatches on the type of the **first argument**, which means we will have to swap the argument order, putting `expr` before `frame`: [c9dec8a6f95e7daf6df897446ce40a7ba4c0a932]

ngjunsiang:
This inconvenience, however, lets us do the following:
https://github.com/nyjc-computing/pseudo-9608/blob/9970442703648344801d5a0b86e08acb47e10c11/pseudocode/resolver.py#L399-L438

This is more verbose, but we can now type the arguments individually. The unittests pass, and `mypy` is happy.

ngjunsiang:
Not forgetting to handle the `passby` keyword argument for `resolveDeclare()`:
https://github.com/nyjc-computing/pseudo-9608/blob/35260df4aa8b285fd51bdb09992be8dd593f942b/pseudocode/resolver.py#L396-L403

ngjunsiang:
Instead of having to pass additional keyword arguments down the chain ... why don't we just merge `resolveDeclare()` into the Declare resolver? [73bf16886e61fee9072bdcf0e52ed35d8837bda2]

Notice that we conveniently don't have to deal with naming each resolving function: `resolve()` serves as a single-dispatch entry point to each resolver.

ngjunsiang:
We might as well do it for all the resolvers then: [f17a4d29fcaaddb212c3564f1d63c7f5000a0ea2]

ngjunsiang:
Naturally we will have to replace calls to each individual resolver with calls to `resolve()`: [bbc115040a8d219a6e107dbdf88d0dfebe5e7ad0], [039c42f3be89023aaa9d29788ccfa3077c98a2aa]

ngjunsiang:
This change has enabled us to cut ~30 lines from the resolver, though that's not the main advantage: besides a slight performance increase, we also avoid having to think of suitable names for each resolver, or to trace the chain of calls. Everything runs through the `resolve()` entry point, aside from procedure calls which still go through `resolveProcCall()` which should only be called through the CallStmt handler.

ngjunsiang:
Let's do the same for the verifiers. First, apply the `singledispatch` decorator and register the verifiers: [6dfa75773c678d5455619d892534a26462a31d69]

ngjunsiang:
Then merge the `verify*()` verifiers into `verify()`: [e10dd0e2bbf7db2d0628557de8afab246b776f17]

ngjunsiang:
Now let's do the same in the interpreter. First, the prep work of swapping `frame` and `expr`/`stmt` args: [605c2d88991df6ce80ad3ea82ee130203ae6fd46]

This is easier to do now that we have unit tests; just keep editing until the tests pass again :v:

ngjunsiang:
Now let's simplify this by merging the evaluators into the singledispatch `evaluate()`. We have a number of evaluators that themselves use the single dispatch technique. Let's apply it to `evalCallable()` first: [b617fe80b1455413d163b83f439960edf15e5234]

One annoying edge case is that `singledispatch`'s `register` decorator seems to throw a wrench into `mypy`'s name-resolving, causing it to be unable to resolve the aliased type definitions in `lang.py` (such as `"Expr"`). So I work around it by unpacking the aliases like so:
https://github.com/nyjc-computing/pseudo-9608/blob/b617fe80b1455413d163b83f439960edf15e5234/pseudocode/interpreter.py#L149

ngjunsiang:
Now applying the same to `evalGet()`: [2c0bef8e099755e8768655da44ae4b8d13d2a41e]

Having Frame inherit from Object is a real pain for typing, since a Frame can contain so many different types while an Object contains a more restricted set of types: [cf9faff828bff718709df267900c0f3241ef17dd]

Very tempting to complain about `mypy` here,  it has pointed out a source of ambiguity in our code that I'll look at refactoring in another chapter. For now, let's narrow the valid return types for Objects and Arrays. We add lots of asserts to silence `mypy` meanwhile: [54ebafa1c7fab6e99892f80f730b08ca7771dbfb]

ngjunsiang:
Hmm .. `evalGet`'s registered dispatchers can actually be merged into `evaluate()`. Might as well do that: [420e6d9d4ccc9848dfe33cb617c906e6f9ecd604]

Cleanup: [b58e7071e4b1f0bd6e2c2ccf883fffe6b23b0d7c]

ngjunsiang:
Finally let's apply single-dispatch to `execute()`: [8b4d6d53f4270e8b10ce24b7603ec63004fa355c]

And merge the executors into `execute()`: [4c241f9548bb7d367996c9e4de72fcbd89d75ce3]

With a notable exception for `execReturn()` which should be explicitly checked for:
https://github.com/nyjc-computing/pseudo-9608/blob/4c241f9548bb7d367996c9e4de72fcbd89d75ce3/pseudocode/interpreter.py#L258-L267
https://github.com/nyjc-computing/pseudo-9608/blob/4c241f9548bb7d367996c9e4de72fcbd89d75ce3/pseudocode/interpreter.py#L280-L282

Since we only expect Return Stmts from Functions, we check for them only in `executeStmts()`, and handle the return value there accordingly (this will typically be passed back to `evalCall()` where the statements were probably executed).

ngjunsiang:
Again we cut about 30 lines in `interpreter.py`, although the main benefit again is not having to name dispatch functions and trace dispatched names.

`mypy` is also better able to help us type-check each dispatch function without a whole bunch of asserts, which we will gradually remove.

ngjunsiang:
~20 lines of asserts removed from `interpreter.py`: [b3b2157cbef094db01143fe1439a0eebeae1affd]

~3 lines of asserts removed from `resolver.py`: [ec443e95f9eff181c4a661f251d1507a329f1b87]

All while keeping `mypy` happy!

PR #85, opened 2022-06-13
28c Shrinkwrap: Compressing loops

ngjunsiang:
**pseudo** is an interpreter, and by its nature contains *a lot* of loops; they're practically unavoidable.

It can be kind of hard to understand what those loops are doing though, so let's explore some techniques to help us with making the code easier to read, through further abstraction.

## Match Until

Our code is full of patterns that look like this:
https://github.com/nyjc-computing/pseudo-9608/blob/728fe3f1660284c2ed0bfbbeac0364434069ad11/pseudocode/parser.py#L386-L387
https://github.com/nyjc-computing/pseudo-9608/blob/728fe3f1660284c2ed0bfbbeac0364434069ad11/pseudocode/parser.py#L391-L394

It's readable, but could use some improvement. Let's start with a helper function:
https://github.com/nyjc-computing/pseudo-9608/blob/e5e9ebf156a36c13789afc27e90ae84217ef1fbb/pseudocode/parser.py#L121-L125

Applying it to the above 4 lines of code, it becomes:
https://github.com/nyjc-computing/pseudo-9608/blob/f6ca67fe4bb6c151fe0864c573b80bc4cb3d18bb/pseudocode/parser.py#L391

With the terminal condition (`'ENDIF'` token) and parsing function placed together, the code becomes easier to understand at a glance as well. If we need to dig into the details of how parsing is done, we'll have to dig into `parseUntilWord()` or `statement5()`, where the details are.

Not all patterns are so readily amenable to the use of `parseUntilWord()`. We'll work through those as well.


ngjunsiang:
The easy ones first: [b2f468dfcf2dcad651d8fba44d4bfdadfd6e7253]
Initial testing caught a bug: our `parseUntilWord()` helper didn't consume the terminal token, so let's do that: [159958fb823d3c51bd4c014c7ff0e00952a7002c]

ngjunsiang:
Now the tricky stuff. First up, CASE statement:
https://github.com/nyjc-computing/pseudo-9608/blob/159958fb823d3c51bd4c014c7ff0e00952a7002c/pseudocode/parser.py#L369-L373

This is coming from the part of `caseStmt()` that parses case values and their respective statements. This isn't something that `parseUntilWord()` can parse in its current form; we don't have a single parsing function that is able to handle this. Let's think about how to make one.

The difficulty here lies in the need to parse a colon-pair (my own term). We also have to parse colon-pairs in two other places: in declarations, and in array indexes. Let's make a `colonPair()` helper function:
https://github.com/nyjc-computing/pseudo-9608/blob/3c599a4ed84494cec25458d531a987fca7c16173/pseudocode/parser.py#L128-L132

And now we can use it in `declare()`: [b3abd322f6d84d17c84ade67822bdeaf547b92ee]

ngjunsiang:
# `colonPair()` in CASE

Okay, let's see how to apply it in `caseStmt()`:
https://github.com/nyjc-computing/pseudo-9608/blob/9f6ff50913154051ad1792d17eb4aaa2f4be19d2/pseudocode/parser.py#L373-L377

First draft:
https://github.com/nyjc-computing/pseudo-9608/blob/83a106781cf19e9672bd3a10ba49855280e9b17d/pseudocode/parser.py#L373-L375

Two questions:
1. How do we apply `parseUntilWord()` to this pattern if it only returns lists?
2. How do we create a function that only takes a `tokens` argument, to pass to `parseUntilWord()`?

In other words, we want code that looks something like (but not necessarily identical to):
`stmts = parseUntilWord(tokens, ['OTHERWISE', 'ENDCASE', valueStmt]`

where `valueStmt()` is our as-yet-unwritten parser.

ngjunsiang:
Let's address question 1 first: as it turns out, you can pass a list of `(key, value)` tuples to `dict()` and it will produce a dict for us. So we can rework the code to:

`stmts = dict(parseUntilWord(tokens, ['OTHERWISE', 'ENDCASE', valueStmt])`

So `parseUntilWord()` should return a list of `(key, value)` tuples, which implies that `valueStmt()` should return a `(caseValue, statement)` tuple:
https://github.com/nyjc-computing/pseudo-9608/blob/753c67acae8e429082715c772b1b6619117d6350/pseudocode/parser.py#L372-L377

ngjunsiang:
# Using lambda functions

Now applying `parseUntilWord()`:
https://github.com/nyjc-computing/pseudo-9608/blob/adc408da877c414b1e1e6e3dc00365c30a520c92/pseudocode/parser.py#L372-L376

Okay, there, we used it. Are we going to have to `def` a new function everywhere we need one? Especially when it's practically a one-liner?

Python lets us create simple anonymous (i.e. unnamed) functions with the `lambda` keyword. Instead of `def valueStmt(tokens): return ...`, we write `lambda tokens: ...`. So we can replace the above code with the following:

https://github.com/nyjc-computing/pseudo-9608/blob/12d3376cc9329cef5975ef51fd850fc27f87edf1/pseudocode/parser.py#L372-L377

ngjunsiang:
`parseWhileWord()` is meant to replace the `while not` loops with which we repeatedly parse statements/expressions until we hit a keyword. There are two more such loops to refactor:
https://github.com/nyjc-computing/pseudo-9608/blob/12d3376cc9329cef5975ef51fd850fc27f87edf1/pseudocode/parser.py#L176-L179
https://github.com/nyjc-computing/pseudo-9608/blob/12d3376cc9329cef5975ef51fd850fc27f87edf1/pseudocode/parser.py#L355-L357

These are ... rather more complex than what a lambda can handle; Python lambdas can only return the evaluated value of a single expression. So we have to do a little more.

ngjunsiang:
Our helpers:
https://github.com/nyjc-computing/pseudo-9608/blob/9183e9e62e401e31035c8bd017d32487f11cc216/pseudocode/parser.py#L135-L147
https://github.com/nyjc-computing/pseudo-9608/blob/9183e9e62e401e31035c8bd017d32487f11cc216/pseudocode/parser.py#L149-L160

And now let's use them. We have a bunch of constructs like these:
```
parse = parser(tokens)
indexes: lang.IndexExpr = (parse(tokens),)
```

ngjunsiang:
Let's use our helpers first.

`buildExprWhileWord()`:
[63412d35edc95706747efd1dbe0218654cba94f4]

`collectExprsWhileWord()`:
[77c2463b4702e4da094571d091db0bc15a66c599]

ngjunsiang:
We also have a whole bunch of `parser()` patterns in our binary expression parsers:
https://github.com/nyjc-computing/pseudo-9608/blob/77c2463b4702e4da094571d091db0bc15a66c599/pseudocode/parser.py#L277-L280

These resist easy compressing, because of the need to pass a `token` and because of that accursed `parser()` pattern. Let's undo the parser dispatch. First re-introduce `value()`: [44ac0e93407cca0d92c65ee4186eb9b685a40172]

And then replace `parser(token)` with `value`: [65b90eb896157dfd5c267c7e71eedba5f0a8026c]

ngjunsiang:
This change benefits us immediately; we can now apply `collectExprsWhileWord()` for index expressions: [4042849e9f8570dff18e63f8dbb50089b56b71f7]

ngjunsiang:
What to do about those Binary Exprs? I feel a headache coming ... we could refactor `builtin.py` and/or `lang.py` to allow us to *not* have to pass a token, but not at this point; we have stable-ish working code and I don't want to do a near-rebuild again.

Let's just use a helper for now:
https://github.com/nyjc-computing/pseudo-9608/blob/c804e7aa168c0f08c63b5b18f1aa607af9582c4f/pseudocode/parser.py#L176-L179

Then we use this helper, with a lambda that helps us extract the `oper`ator and `right` expression: [c804e7aa168c0f08c63b5b18f1aa607af9582c4f]

ngjunsiang:
Oops, I snoozed this project too long and can't remember where I was going with this. Let's close this chapter here; it's enough shrinkwrapping I think for me to move on to the next feature.

PR #86, opened 2022-10-17
29 Cleanup

ngjunsiang:
After a long hiatus, let's close some outstanding issues and get the code to working condition again.

First fixes: our helper functions for parsing used a `parserMap` with the token words (e.g. `[`, `.`) as keys. When we attempt to retrieve the parser using the token, we therefore need to use the token word instead of the entire token: [255c26a9d20fda82c0d029d1cf1fd67bd68bf039]


ngjunsiang:
Our `TypeStmt` stored a list of DeclareStmts instead of a list of Declares, that is now fixed: [bde94bb7999760c322905ec9308121465592629c]

ngjunsiang:
`buildExprWhileWord` was not consuming the `[` for IndexExprs, causing the parser to choke. So we add a keyword argument for index exprs to consume the `[` before parsing the rest of the tokens: [a76675918b4c7450efa5c5088e98d17dd1251270]

https://github.com/nyjc-computing/pseudo-9608/blob/a76675918b4c7450efa5c5088e98d17dd1251270/pseudocode/parser.py#L135-L151

ngjunsiang:
The typing for sequences of Exprs, Stmts, etc is a mess. Let's add some conveniences: [cfbc8cde82f2f1ee20bcb247e9e9bebc65a6332c]

ngjunsiang:
Our `parseUntilWord` helper forgot that sometimes we need to consume the word, and sometimes we don't. So let's split that into two helpers:

https://github.com/nyjc-computing/pseudo-9608/blob/016af58aebb259f2868cd1c497bfe2aa64079862/pseudocode/parser.py#L121-L148

And then update all the helper calls to use the correct version: [016af58aebb259f2868cd1c497bfe2aa64079862]

ngjunsiang:
# Parser fixes

Now, the parser. First, let's use the convenience types we created earlier: [abf4bf2df3006e62ad629a4fefdc94584c35e3b3]

ngjunsiang:
`resolveNames` is a mess because it took on too much, trying to handle Names in both Exprs and Stmts. That is easily fixed: from now on, it is only responsible for taking in an UnresolvedName, and returning a GetName.

Its other responsibilities are now moved into `resolveNamesInExpr`, which we rename to `resolveNamesInTarget` to better reflect its flexibility in handling both Expr and Stmt: [9d02adf9509e5ed6ffaedb748f690565cf303b0b]

ngjunsiang:
Not forgetting to rename all the relevant function calls: [4fca207389425c0c5709c1600d1e44631bee5c2b]

ngjunsiang:
Now, an actual fix: we currently resolve the parameters and statements in callables (Functions and Procedures) using the global frame ... but both are being executed in the local context, so they should be resolved using the local frame. Let's fix that: [09db8d6940e7d16d1e2e661e5dedd11a9eac4562]

ngjunsiang:
# Interpreter fixes

I came to hate multiline function definitions, so let's turn them back to single-line defs where possible: [2fbb14392e14d9fb4a28999d207ddbef2541050f]

ngjunsiang:
Just as we resolve functions/procedures with their local frames rather than the global, we also interpret them likewise: [15d1b19b201707abc0ff520612c1ce8da859e0b0]

ngjunsiang:
Now we have a problem: Builtin functions (system functions) do not have a frame ... yet. Let's give them one: [86b319f2bfa2f337e496d20925466977ac747885]

ngjunsiang:
A very irritating problem: Python uses `args` to reference arguments passed to a function. So we have to rename all uses of that variable: [fea92ee922cb94240b55b3724bee6d3f93065ef9]

ngjunsiang:
Now we're stuck: how do we resolve the system frame with the global frame?

https://github.com/nyjc-computing/pseudo-9608/blob/fea92ee922cb94240b55b3724bee6d3f93065ef9/pseudocode/system.py#L21-L28

Our system frame is only initialised once, and then imported in `__init__`:

https://github.com/nyjc-computing/pseudo-9608/blob/fea92ee922cb94240b55b3724bee6d3f93065ef9/pseudocode/system.py#L17

https://github.com/nyjc-computing/pseudo-9608/blob/fea92ee922cb94240b55b3724bee6d3f93065ef9/pseudocode/__init__.py#L22

ngjunsiang:
We are gonna need to rearrange lots of stuff. Let's start by namespacing things properly: [44bf8908dfd87dfc9288b344e36552989667d9ae]

ngjunsiang:
Now we need to somehow pass the global frame back to the `sysFrame` to be used in the system functions. We need to separate the function declarations from the function definitions. After this change, our `__init__.py` would looks like this:
https://github.com/nyjc-computing/pseudo-9608/blob/cc0a87a3f8d3f884434f37e9240da09382fe28a5/pseudocode/__init__.py#L94-L101

And our `system.initFrame()` is split up; `initFrame` only declares the functions now, and the functions are defined in `resolveGlobal()` to hint at where they should be defined:
https://github.com/nyjc-computing/pseudo-9608/blob/cc0a87a3f8d3f884434f37e9240da09382fe28a5/pseudocode/system.py#L40-L72

This points at a key distinction between declaration and definition: declaration reserves the names in the frame, definition is what maps the code to the name.

ngjunsiang:
I foresee some pain here ... we'll definitely be adding more system functions, and each addition will involve:
- a function definition in Python (using `def`)
- a function declaration in the system frame
- a function definition in Pseudo (using `lang.Builtin`)

That's three places to edit. Let's declare and define the system functions in a loop to reduce the likelihood of errors. We put the function return and parameter types in a data structure — just a simple list of tuples:
https://github.com/nyjc-computing/pseudo-9608/blob/0ed41510b0483fab10397c708eff2ed8b9ca1889/pseudocode/system.py#L40-L47

And now it's child's play to loop through that and declare/define the functions:
https://github.com/nyjc-computing/pseudo-9608/blob/0ed41510b0483fab10397c708eff2ed8b9ca1889/pseudocode/system.py#L49-L69

PR #89, opened 2022-10-18
29b Trivialities

ngjunsiang:
I'm using this chapter to make changes that don't really fit anywhere else, in preparation for implementing new features.

First up: adding a whole bunch of system functions: [5a49ee7e96029fcc4c03759863354c97539afd08], [5c6d805cd7ed85390a442ebdd6cec74d72ab8abc]
- LENGTH
- LEFT
- RIGHT
- INT
- MOD
- MID
- LCASE
- DIV
- INT_TO_STRING
- REAL_TO_STRING

These functions are mentioned in the appendix of 9608 syllabus exam papers, so I might as well make them available.


ngjunsiang:
The first major test of the interpreter will be to get an algorithm working: let's start with bubblesort.

https://github.com/nyjc-computing/pseudo-9608/blob/7d24706f90953bab9acd8ea9a16110dd0788ee84/main.pseudo

ngjunsiang:
A minor fix: the column numbers of error reports are always smaller by one.

https://github.com/nyjc-computing/pseudo-9608/pull/89/commits/7f4f4d08ff51b0d76a8d0a7083fa8a76c1ea7f73

ngjunsiang:
And an error-reporting fix: the column numbers of error reports are always smaller by one. Not anymore!

https://github.com/nyjc-computing/pseudo-9608/pull/89/commits/c9a0a6e7a981cfa1b9906f594b3cde11f82ae30d

We also need to adjust the cursor position in error reports after this:

https://github.com/nyjc-computing/pseudo-9608/pull/89/commits/202c2c74cc99cced889faf006a22c324e1226043

ngjunsiang:
The current interpreter only lets us use values and not expressions in GetIndex (e.g. `Data[1]` but not `Data[i + 1]`). Let's fix that: [418e710eec6f533317a47aea7e58358785e92411]

ngjunsiang:
```
$ pseudo main.pseudo 
[Line 5] 
         ^
ParseError: <[Line 5 column 1] {None} '\n'>: Unrecognised token
```

Funny bug: our interpreter can't handle multiple line breaks!

I made a `skipBlankLines()` helper and was scattering it throughout `parser.py` when it struck me: this shouldn't be the parser's job! There is no inherent meaning in multiple line breaks, and we should be getting rid of the excess ones in ... the scanner. Let's do that: [https://github.com/nyjc-computing/pseudo-9608/pull/89/commits/5b9300bb8895c947a6afc2867506a89e1852fc53]

ngjunsiang:
```
$ pseudo main.pseudo 
[Line 7]         FOR j <- 1 TO 10 - i
                 ^
ParseError: <[Line 7 column 9] {None} 'FOR'>: Unrecognised token
```

Huh weird, why isn't the interpreter recognising a FOR? Digging into our forStmt()` parser:
https://github.com/nyjc-computing/pseudo-9608/blob/5b9300bb8895c947a6afc2867506a89e1852fc53/pseudocode/parser.py#L518-L520

This uses `statement5()` to parse statements between FOR and EDNFOR.
https://github.com/nyjc-computing/pseudo-9608/blob/5b9300bb8895c947a6afc2867506a89e1852fc53/pseudocode/parser.py#L645-L661

Ohh ... `statement5()` is already lower precedence than FOR, so it never gets recognised! To enable IF/FOR/WHILE/REPEAT nesting, we need to use the `statement4()` parser instead: [574fbb7cdd0f65a453c65f044ebf289d33c13bc5]

ngjunsiang:
An annoying bug helped me realise that token representation could be improved. Some tokens have value `None`, and this looks weird when printed. So let's make it optional:
https://github.com/nyjc-computing/pseudo-9608/blob/130fb53a7cb0e3c8d0984c37ec9f868a04628afd/pseudocode/lang.py#L83-L86

Expr has a `token` property while Name is still behind the times, using a `token()` method. Time to update it: [130fb53a7cb0e3c8d0984c37ec9f868a04628afd]

ngjunsiang:
```
$ pseudo main.pseudo 
[Line 8]             IF Data[j] > Data[j + 1]
                        ^
LogicError: [Line 8 column 16] 'Data' : Undeclared
```

Okay ... why is Data not being detected in the Procedure? Looks like we need to move the declaration up to before the procedure definition: [696d8d6344ba1bad988c342f87e18de950d132b9]

```
$ pseudo main.pseudo 
[Line 33] FOR i <- 1 TO 10
              ^
LogicError: [Line 33 column 5] 'i' : Undeclared
```
Okay, now what ... looks like we were missing a declaration in the main program. Our interpreter is doing the correct thing after all!
Let's fix that: [6bbd9be6174061cb7cc60c797f2f1c3865f9fecd]
And also improve the output: [9bc10f2575e1f58bc82261decce8a98d6442cc23]

ngjunsiang:
Let's try something interesting: Dynamically determining the loop end with a variable: [0b57de56cb5b6d782208e381b8c9c6d1867dd833]

https://github.com/nyjc-computing/pseudo-9608/blob/0b57de56cb5b6d782208e381b8c9c6d1867dd833/main.pseudo#L1-L39

ngjunsiang:
```
$ pseudo main.pseudo 
[Line 12]     FOR i <- 1 TO Length - 1
                                   ^
ParseError: [Line 12 column 26] '-' <function sub at 0x7f53b9eb94c0>: Expected ('\n',) at end of FOR
```

Looks like our parser expects only a single value for the start and end. Let's extend it to allow expressions: [b7e021ca5631ceef46e1195e24b271279b846ec0]

I'll use the chance to make a nicer-looking EOF token too. It is represented with the value `\0` in many languages, so let's use that: [859c7eece64dec832a6c6a3b1724c7d2f86cfb7b]

ngjunsiang:
Lastly, there's a lot of stuff going on in the Resolver for Procedure and Function statements. Let's rearrange things a bit and add comments to make reading easier: [31f8c697ccaff3ec73f60d4fd66195e654244ce3]

ngjunsiang:
We might as well add this as another test in our test suite: [980dc2b8ff3ac5a363474974a4c34e5ab0b659c1]

ngjunsiang:
A long-overdue change: Ever since we started using Assign Exprs instead of only AssignStmts, loops didn't really need an AssignStmt for the init anymore. Let's use an Assign Expr instead: [7f7704c7cf2344d0b951239c03672f0dd363c2da]

And ... tests passed!

ngjunsiang:
Let's try an insertion sort:
```
DECLARE Length : INTEGER
DECLARE Data : ARRAY[1:10] OF INTEGER
DECLARE i : INTEGER

Length <- 10

PROCEDURE InsertSort(BYREF Data: ARRAY[1:10] OF INTEGER, Length: INTEGER)
    DECLARE i : INTEGER
    DECLARE p : INTEGER
    DECLARE t : INTEGER
    DECLARE Temp : INTEGER

    IF Length > 1
      THEN
        FOR i <- 2 TO Length
            p <- 1
            WHILE p < i AND Data[i] > Data[p] DO
                p <- p + 1
            ENDWHILE
            Temp <- Data[i]
            WHILE i > p DO
                Data[i] <- Data[i - 1]
                i <- i - 1
            ENDWHILE
            Data[i] <- Temp
        ENDFOR
    ENDIF
ENDPROCEDURE

Data[1] <- 7
Data[2] <- 9
Data[3] <- 4
Data[4] <- 1
Data[5] <- 2
Data[6] <- 3
Data[7] <- 5
Data[8] <- 10
Data[9] <- 6
Data[10] <- 8

CALL InsertSort(Data, Length)

FOR i <- 1 TO 10
    OUTPUT i, ": ", Data[i]
ENDFOR
```

```
$ pseudo main.pseudo 
1: 1
2: 2
3: 3
4: 4
5: 5
6: 6
7: 7
8: 8
9: 9
10: 10
```

It works too 😎 Let's add it to our test suite as well: [f39ee85ceffeb53eed98099b1b94b4f9f9f8b156]

PR #92, opened 2022-10-18
30b New features: concatenation with & operator

ngjunsiang:
So we managed to add comments (`//`) without touching anything beyond the scanner. Let’s work on something that extends beyond the scanner.

In the exam appendix, 9608 pseudocode uses `&` as the concatenation operator for strings. `String1 & String2` will return the result of concatenating `String1` and `String2`.

# Changes in Builtins

Let’s start by adding recognition for this operator in Pseudo’s symbols:
https://github.com/nyjc-computing/pseudo-9608/blob/c1a609d3d5bf6bd1600a0f002f420d9a7e366d7a/pseudocode/builtin.py#L134

The function associated with this is `concat()`, which we will need to write:
https://github.com/nyjc-computing/pseudo-9608/blob/c1a609d3d5bf6bd1600a0f002f420d9a7e366d7a/pseudocode/builtin.py#L82-L83
It looks *suspiciously* identical to `add()`, and it is! Because that's how Python does it. But I like to keep them as separate functions, because 9608 pseudocode does not overload functions in any way and I like to keep each function’s purpose and name separate.

Lastly, let’s add `concat()` to the OPERATORS dictionary so the parser can look it up and associate it with the `&` symbol:
https://github.com/nyjc-computing/pseudo-9608/blob/c1a609d3d5bf6bd1600a0f002f420d9a7e366d7a/pseudocode/builtin.py#L117-L132

Full commit: [c1a609d3d5bf6bd1600a0f002f420d9a7e366d7a]


ngjunsiang:
# Changes in Scanner

None! Our scanner, using the info in `builtin.py`, will successfully detect `&` as a single-symbol token, without any further changes needed.

ngjunsiang:
# Changes in Parser

The `&` token, when encountered, should be parsed as a binary expression. So the changes needed in our parser should only be confined to the binary-parsing helper.

We add one more symbol for it to recognise in its parsing:

https://github.com/nyjc-computing/pseudo-9608/blob/c7685a80ebcab24e2ed7e85230129355f55fe7bb/pseudocode/parser.py#L322-L330

In other languages that have a separate concatenation operator, you might have numbers and strings mixed in a single expression, and concatenation usually has lower precedence than addition/subtraction. In 9608 pseudocode, an expression can only have numbers or strings, so we can put concatenation at the same precedence as addition/subtraction.

Full commit: [c7685a80ebcab24e2ed7e85230129355f55fe7bb]

ngjunsiang:
# Changes in Resolver

On to the resolver. We need to typecheck concatenation expressions under the binary expression resolver, so let's add a separate case for it:
https://github.com/nyjc-computing/pseudo-9608/blob/380df6ac5a8e24db846fdd31dc85bb95289569e2/pseudocode/resolver.py#L257-L264

I used an `in` membership check instead of an `is` identity check, because I suspect I'll need to add mroe cases here once I work on the CHAR data type.

Full commit: [380df6ac5a8e24db846fdd31dc85bb95289569e2]

ngjunsiang:
# Changes in Interpreter

Our interpreter already handles all binary expressions the same way, evaluating the left and right operands and then passing them to the operator function. That should handle concatenation just fine too.

Current code:
https://github.com/nyjc-computing/pseudo-9608/blob/380df6ac5a8e24db846fdd31dc85bb95289569e2/pseudocode/interpreter.py#L92-L95

ngjunsiang:
Time to test these changes out. Let's change our output line to use string concatenation instead of relying on OUTPUT's own conversion: [8868eb031beb0ba264856af566380b7ebf0fdea8]

```
// A comment
DECLARE Length : INTEGER
DECLARE Data : ARRAY[1:10] OF INTEGER
DECLARE i : INTEGER

Length <- 10

PROCEDURE InsertSort(BYREF Data: ARRAY[1:10] OF INTEGER, Length: INTEGER)
    DECLARE i : INTEGER
    DECLARE p : INTEGER
    DECLARE t : INTEGER
    DECLARE Temp : INTEGER

    IF Length > 1
      THEN
        FOR i <- 2 TO Length
            p <- 1
            WHILE p < i AND Data[i] > Data[p] DO
                p <- p + 1
            ENDWHILE
            Temp <- Data[i]
            WHILE i > p DO
                Data[i] <- Data[i - 1]
                i <- i - 1
            ENDWHILE
            Data[i] <- Temp
        ENDFOR
    ENDIF
ENDPROCEDURE

Data[1] <- 7
Data[2] <- 9
Data[3] <- 4
Data[4] <- 1
Data[5] <- 2
Data[6] <- 3
Data[7] <- 5
Data[8] <- 10
Data[9] <- 6
Data[10] <- 8

CALL InsertSort(Data, Length)

FOR i <- 1 TO 10
    OUTPUT "Index: " & INT_TO_STRING(i) & ": " & INT_TO_STRING(Data[i])
ENDFOR
// End of code
```

ngjunsiang:
```
  File "__init__.py", line 132, in run
    tokens, lines = scanner.scan(src)
  File "scanner.py", line 179, in scan
    raise builtin.ParseError(
  File "builtin.py", line 15, in __init__
    self.line = token.line
AttributeError: 'str' object has no attribute 'line'
```

Oh dear ... Pseudo does not like our use of the `_` character in the function `INT_TO_STRING()`. This is kind of funny, because `INT_TO_STRING` comes from the appendix of the 9608 exam, yet the 9608 pseudocode guide says "accented letters
and other characters, **including the underscore**, should not be used."

We are going to have to make a decision here. The pseudocode guide says "keywords are in uppercase", and also gives some examples of system-provided functions (e.g. `RANDOMBETWEEN`). I can think of two options here:

1. Allow `_` in identifiers
2. Rename system functions to remove the underscore.

I'm going to go the lazy route and take option 2: [2a6122f71264470b9e855753b537955ad666bf7d]

ngjunsiang:
One more silly typo fix: [81fa2eb6c7a3563033dd38a047b71a187bd6f5df]

```
$ pseudo main.pseudo 
Index: 1: 1
Index: 2: 2
Index: 3: 3
Index: 4: 4
Index: 5: 5
Index: 6: 6
Index: 7: 7
Index: 8: 8
Index: 9: 9
Index: 10: 10
```

And now it works!

ngjunsiang:
# Final steps

Let's add a test for that: [068f535254440c345dde7b860f4fd0fc8ca94e8a]

And then some cleanup: [2bfad08dafdd97e3e6885a96c5b06805ee016c9a]

And we're done for this chapter.

PR #93, opened 2022-10-19
30c New features: Fixing CASE evaluations

ngjunsiang:
We've implemented two new features so far, with few hiccups. Let's try something hairier: fixing a bug.

In [`main.pseudo`](https://github.com/nyjc-computing/pseudo-9608/blob/fc4307d332b941985f0d9ef9643d9ef6458ab6e6/main.pseudo):
```
DECLARE Test : STRING

Test <- "Hi"

CASE OF Test
  "Hello": OUTPUT "Oh, hello there!"
  "Hi": OUTPUT "Hihi!"
  "Hi": OUTPUT "Hi again!"
  "Yo": OUTPUT "Wassup?"
  0: OUTPUT "Huh?"
  0: OUTPUT "Sense of deja vu ..."
  1: OUTPUT "No way..."
ENDCASE
```

```
$ pseudo main.pseudo 
Hi again!
```

Doesn't seem like anything is wrong here ... the code runs, right? But this should rightfully throw an error or two:

1. The case values `"Hi"` and `0` are repeated, and our parser/resolver should have pointed that out since it is probably an inadvertent error on the user's part if it happens (unlike in this case).
2. The case condition, `Test`, is declared as STRING type, so all case values should also be STRING. `0` and `1` are invalid case values, and the resolver should have pointed this out.

Let's make that happen.

### DIFF
--- a/main.pseudo
+++ b/main.pseudo
@@ -1,47 +1,13 @@
-// A comment
-DECLARE Length : INTEGER
-DECLARE Data : ARRAY[1:10] OF INTEGER
-DECLARE i : INTEGER
+DECLARE Test : STRING
 
-Length <- 10
+Test <- "Hi"
 
-PROCEDURE InsertSort(BYREF Data: ARRAY[1:10] OF INTEGER, Length: INTEGER)
-    DECLARE i : INTEGER
-    DECLARE p : INTEGER
-    DECLARE t : INTEGER
-    DECLARE Temp : INTEGER
-
-    IF Length > 1
-      THEN
-        FOR i <- 2 TO Length
-            p <- 1
-            WHILE p < i AND Data[i] > Data[p] DO
-                p <- p + 1
-            ENDWHILE
-            Temp <- Data[i]
-            WHILE i > p DO
-                Data[i] <- Data[i - 1]
-                i <- i - 1
-            ENDWHILE
-            Data[i] <- Temp
-        ENDFOR
-    ENDIF
-ENDPROCEDURE
-
-Data[1] <- 7
-Data[2] <- 9
-Data[3] <- 4
-Data[4] <- 1
-Data[5] <- 2
-Data[6] <- 3
-Data[7] <- 5
-Data[8] <- 10
-Data[9] <- 6
-Data[10] <- 8
-
-CALL InsertSort(Data, Length)
-
-FOR i <- 1 TO 10
-    OUTPUT "Index: " & INTTOSTRING(i) & ": " & INTTOSTRING(Data[i])
-ENDFOR
-// End of code
+CASE OF Test
+  "Hello": OUTPUT "Oh, hello there!"
+  "Hi": OUTPUT "Hihi!"
+  "Hi": OUTPUT "Hi again!"
+  "Yo": OUTPUT "Wassup?"
+  0: OUTPUT "Huh?"
+  0: OUTPUT "Sense of deja vu ..."
+  1: OUTPUT "No way..."
+ENDCASE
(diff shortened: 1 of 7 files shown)


ngjunsiang:
# Changes to Lang

Thinking this through, to retain type information of case values, we would need to store the Literal Expr instead of just its value. Let's make those changes in the language:
https://github.com/nyjc-computing/pseudo-9608/blob/eef5395e4c25dc52602983ab5beb76b2d8135e1c/pseudocode/lang.py#L55

I also use the chance to clean up the type organisation, full commit: [eef5395e4c25dc52602983ab5beb76b2d8135e1c]

ngjunsiang:
# Changes to Parser

Now our parser should pass the entire Literal instead of just its value: [d4f0fc7ddd30a5d6eb8b7dfbdf11cf26c7b585f9]

https://github.com/nyjc-computing/pseudo-9608/blob/d4f0fc7ddd30a5d6eb8b7dfbdf11cf26c7b585f9/pseudocode/parser.py#L464-L470

ngjunsiang:
# Changes to Resolver

Now our resolver has the information to carry out typechecks on the case values, to ensure they have the same time as the condition's type: [d0e7e625a694296a83d77451cb711a86a1950a89]

https://github.com/nyjc-computing/pseudo-9608/blob/d0e7e625a694296a83d77451cb711a86a1950a89/pseudocode/resolver.py#L392-L396

ngjunsiang:
# Changes to Interpreter

Now we can't use a simple `if cond in stmt.stmtMap` check anymore; we'll need to evaluate the Literal to compare values. We might figure out some way to optimise this in future but not right now.

After the changes: [c87e5b5af0ed7255992af185af81abce3c510aaa]

https://github.com/nyjc-computing/pseudo-9608/blob/c87e5b5af0ed7255992af185af81abce3c510aaa/pseudocode/interpreter.py#L255-L262

This relies on a seldom-used feature in Python: a `for-else` loop. The `else` block executes only if the `for` loop completes normally, i.e. was not terminated with `break`.

ngjunsiang:
Now we run into a fundamental problem: Python can't hash a Literal, because it is not a [hashable](https://docs.python.org/3/glossary.html#term-hashable)!

Python needs a hashable key for dict, because dict is a hashtable, and needs a way to convert the key to an integer index. We'll implement that in our Literal class, make it hashable by implementing the `__hash__` and `__eq__` dunder methods: [713b9b1adf0833e4c00720c15a552d9ca78ac24f]

https://github.com/nyjc-computing/pseudo-9608/blob/713b9b1adf0833e4c00720c15a552d9ca78ac24f/pseudocode/lang.py#L552-L566

In the parser, instead of taking the parsed `caseValue: statement` colon pairs and turning them into a dictionary directly, we now do it in a loop, checking for existing values before we add:

https://github.com/nyjc-computing/pseudo-9608/blob/713b9b1adf0833e4c00720c15a552d9ca78ac24f/pseudocode/parser.py#L464-L475

ngjunsiang:
And now it works:

```
$ pseudo main.pseudo 
[Line 8]   "Hi": OUTPUT "Hi again!"
           ^
ParseError: [Line 8 column 3] '"Hi"' Hi: Repeated CASE value
```

Ugly token representation though. We might as well take the chance to improve it. We don't need both the value and the word displayed, since one is usually sufficient; if there isn't a value, then we use the word: [a5fbac50c9d315716960b080f3e5b91d04cd4620]

```
$ pseudo main.pseudo 
[Line 8]   "Hi": OUTPUT "Hi again!"
           ^
ParseError: [Line 8 column 3] 'Hi': Repeated CASE value
```

Much better.

ngjunsiang:
```
python -m unittest
...............
Ran 15 tests in 0.389s

OK
```

Unit tests passed, we're almost good to go!

Let's add this as a test case too: [6488892d1b9fdee7ce76c6e96b348f0f49b1df92]

And we are done here.

PR #96, opened 2022-10-22
31a Defining behaviour: Cleaning up (again)

ngjunsiang:
Last chapter we started adding new features just to see what is involved. And we fixed a bug, and uncovered a bunch of issues while trying to please `mypy`.

This chapter is all about defining what Pseudo should and should not do, and getting our Python types to make that easier for us.

We start with a general syntax cleanup in `lang.py`, which is going to be the crux of this chapter. No changes, just making things look nicer, easier on the eye, and less painful to read.

### DIFF
--- a/pseudocode/interpreter.py
+++ b/pseudocode/interpreter.py
@@ -263,11 +263,14 @@ def _(stmt: lang.Case, frame: lang.Frame, **kwargs) -> None:
 
 @execute.register
 def _(stmt: lang.If, frame: lang.Frame, **kwargs) -> None:
-    cond = evaluate(stmt.cond, frame)
-    if cond in stmt.stmtMap:
-        executeStmts(stmt.stmtMap[cond], frame, **kwargs)
-    elif stmt.fallback:
-        executeStmts(stmt.fallback, frame, **kwargs)
+    condValue = evaluate(stmt.cond, frame)
+    for caseValue, stmts in stmt.stmtMap.items():
+        if evaluate(caseValue, frame) == condValue:
+            executeStmts(stmts, frame, **kwargs)
+            break
+    else:  # for loop completed normally, i.e. no matching cases
+        if stmt.fallback:
+            executeStmts(stmt.fallback, frame, **kwargs)
     
 @execute.register
 def _(stmt: lang.While, frame: lang.Frame, **kwargs) -> None:
--- a/pseudocode/interpreter.py
+++ b/pseudocode/interpreter.py
@@ -265,7 +265,7 @@ def _(stmt: lang.Case, frame: lang.Frame, **kwargs) -> None:
 def _(stmt: lang.If, frame: lang.Frame, **kwargs) -> None:
     cond = evaluate(stmt.cond, frame)
     if cond in stmt.stmtMap:
-        executeStmts(stmt.stmtMap[True], frame, **kwargs)
+        executeStmts(stmt.stmtMap[cond], frame, **kwargs)
     elif stmt.fallback:
         executeStmts(stmt.fallback, frame, **kwargs)
     
--- a/pseudocode/lang.py
+++ b/pseudocode/lang.py
@@ -37,7 +37,7 @@ Type = str  # pseudocode type, whether built-in or declared
 IndexDigit = Union["Expr"]
 IndexExpr = Tuple[IndexDigit, ...]  # Array indexes
 IndexRange = Tuple[int, int]  # Array ranges (start, end)
-Passby = LiteralType['BYREF', 'BYVALUE']
+Passby = LiteralType["BYREF", "BYVALUE"]
 
 # Plurals
 Exprs = Iterable["Expr"]
@@ -80,7 +80,7 @@ class Token:
     and Stmts.
     It also encapsulates code information for error reporting.
     """
-    __slots__ = ('line', 'column', 'type', 'word', 'value')
+    __slots__ = ("line", "column", "type", "word", "value")
     line: int
     column: int
     type: Type
@@ -97,14 +97,14 @@ class Name:
     """Name represents a meaningful name, either a custom type or a
     variable name.
     """
-    __slots__ = ('name', '_token')
+    __slots__ = ("name", "_token")
 
     def __init__(self, name: NameKey, *, token: "Token") -> None:
         self.name = name
         self._token = token
 
     def __repr__(self) -> str:
-        return f'Name({self.name})'
+        return f"Name({self.name})"
 
     def __str__(self) -> NameKey:
         return self.name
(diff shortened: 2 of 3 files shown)


ngjunsiang:
First thing to clarify:

https://github.com/nyjc-computing/pseudo-9608/blob/ce9fe63429b40fda09894968f7445bf530a9f71e/pseudocode/parser.py#L260-L286

These are Exprs involving Names. The parser picks up names (through the `identifier()` parser helper), turning them into UnresolvedNames, which the resolver later replaces with GetNames.

`mypy` is extremely unhappy with this, and it takes a lot of `assert` statements to please it here. We need a type that covers "Exprs that can evaluate to Arrays, Objects, or Functions".

But first let's fix the existing type problems. I won't explain in detail here, as I have done elsewhere, since these for the most part do not add to the purpose of this chapter.

ngjunsiang:
Except for one interesting fix: [6c54b45b74af57220a0d118458cbe0d3d78246f4]

Explained in this Stack Overflow link: https://stackoverflow.com/a/54816069

I typed `Literal.__eq__` to only allow checks against other Literals, but Python wants this dunder to accept all objects, so that it can try calling `other.__eq__(self)` if `other` implements `__req__`. Fine.

ngjunsiang:
Oops, I broke the unit tests somewhere along the way. Ah, there it is: our parser started giving Literals to Case and If, but we only fixed the interpreter for Case. Now we fix it for If too: [d17fc040908c7b24a5007815d821cc3851609440]

ngjunsiang:
Back to the problem. We need a way to tell `callExpr()`, `attrExpr()`, and `indexExpr()` that GetNames, GetIndexs, and GetAttrs are all fine to accept. We actually do have one:

https://github.com/nyjc-computing/pseudo-9608/blob/6c54b45b74af57220a0d118458cbe0d3d78246f4/pseudocode/lang.py#L64-L65

But I polluted it with UnresolvedName. Let's remove that—no, let’s do better: Let's have all 3 GetExpr subtypes inherit from a GetExpr parent type: [23b3b3226cbc52b211d95deec1b117d5cfdde544]

https://github.com/nyjc-computing/pseudo-9608/blob/23b3b3226cbc52b211d95deec1b117d5cfdde544/pseudocode/lang.py#L620-L664

ngjunsiang:
This is bound to make `mypy` complain, but we ahve to break some things first to build them back better. Still, this should not be causing unit tests to fail. Quick check:

```
$ python -m unittest
.................
Ran 17 tests in 0.120s

OK
$ python -m mypy pseudocode/*.py
pseudocode/parser.py:276: error: Argument 1 to "GetAttr" has incompatible type "Union[UnresolvedName, Call, GetAttr, GetIndex]"; expected "Union[GetExpr, Call]"
pseudocode/parser.py:286: error: Argument 1 to "GetIndex" has incompatible type "Union[UnresolvedName, Call, GetAttr, GetIndex]"; expected "Union[GetExpr, Call]"
pseudocode/parser.py:447: error: Argument 1 to "Input" has incompatible type "UnresolvedName"; expected "GetExpr"
pseudocode/parser.py:657: error: Argument 2 to "ReadFile" has incompatible type "UnresolvedName"; expected "GetExpr"
Found 4 errors in 1 file (checked 8 source files)
```

Phew, okay.

ngjunsiang:
The easy fixes first: we can now make it clearer that `indexExpr()` and `attrExpr(0` parsers expect a GetExpr: [66509cbf1dc8d340fbe5a0447b8e47a9e90bf1bc]

ngjunsiang:
Now the tricky part: Input and ReadFile expect a variable to put user/file input into. This could be a variable name, but it could also be an indexed array slot or object attribute slot. So they rightfully expect a GetExpr, but instead could be getting an UnresolvedName:

https://github.com/nyjc-computing/pseudo-9608/blob/66509cbf1dc8d340fbe5a0447b8e47a9e90bf1bc/pseudocode/parser.py#L444-L447

https://github.com/nyjc-computing/pseudo-9608/blob/66509cbf1dc8d340fbe5a0447b8e47a9e90bf1bc/pseudocode/parser.py#L652-L657

Time to update the code.

ngjunsiang:
Hmm, what we want is an assignment target. We could borrow some code from `assignment()`:

https://github.com/nyjc-computing/pseudo-9608/blob/66509cbf1dc8d340fbe5a0447b8e47a9e90bf1bc/pseudocode/parser.py#L414-L424

That looks pretty unwieldy though. Anytime we start to repeat code, it's time to ... add another helper!

https://github.com/nyjc-computing/pseudo-9608/blob/40755d3b7c5b053afa64873d14771a8fb7decab7/pseudocode/parser.py#L289-L296

Our `target()` helper takes a starting token—an UnresolvedName—and calls `indexExpr()` or `attrExpr()` as appropriate, aiming to produce a valid assignment target.

Because this is meant to be used to parse assignment targets only, we **do not** use it in `nameExpr()`, which is meant to produce GetExprs:

https://github.com/nyjc-computing/pseudo-9608/blob/40755d3b7c5b053afa64873d14771a8fb7decab7/pseudocode/parser.py#L299-L314

The difference is assignment targets go on the left hand side of an assignment (`<-`), while GetExprs go on the right hand side.

ngjunsiang:
Now we make those changes. First, let's clean up that confusing mess of types. In place of NameExprs and NameKeyExprs and all that old stuff, let's use SetExprs and GetExprs, and also versions that accept UnresolvedNames:
https://github.com/nyjc-computing/pseudo-9608/blob/98815ced92b97c10916c87aeec7458ed21391fc9/pseudocode/lang.py#L62-L71

Much of this commit is just having `lang.py` and `parser.py` use these new types. The main changes are in our `assignment()`, `inputStmt()`, and `readFileStmt()` helpers, which nwo use the new `target()` helper:
https://github.com/nyjc-computing/pseudo-9608/blob/98815ced92b97c10916c87aeec7458ed21391fc9/pseudocode/parser.py#L423-L428
https://github.com/nyjc-computing/pseudo-9608/blob/98815ced92b97c10916c87aeec7458ed21391fc9/pseudocode/parser.py#L442-L446
https://github.com/nyjc-computing/pseudo-9608/blob/98815ced92b97c10916c87aeec7458ed21391fc9/pseudocode/parser.py#L651-L657

Full commit: [98815ced92b97c10916c87aeec7458ed21391fc9]

ngjunsiang:
```
$ python -m mypy pseudocode/*.py
Success: no issues found in 8 source files
$ python -m unittest
.................
Ran 17 tests in 0.195s

OK
```

`mypy` is happy, unit tests pass, looks like everything's peachy so far.

PR #97, opened 2022-10-24
31b Defining behaviour: EOF handling and Returns

ngjunsiang:
Having cleaned up GetExprs and SetExprs, let's tackle EOFs.

Right now, the EOF token is added in the parser:
https://github.com/nyjc-computing/pseudo-9608/blob/9c171ab7b051932752d7c14c3a513434704e8002/pseudocode/parser.py#L754-L758

I don't like this. When we split responsibilities, the scanner is supposed to add tokens, and the parser is supposed to parse them. Let's move this to the scanner instead: [e153dca6a0a0b6f60654ad3b2bc4355ba3c5fc34]

### DIFF
--- a/main.pseudo
+++ b/main.pseudo
@@ -7,8 +7,7 @@ FUNCTION Respond(Prompt: STRING) RETURNS STRING
       "Hello": RETURN "Oh, hello there!"
       "Hi": RETURN "Hihi!"
       "Yo": RETURN "Wassup?"
-      0: RETURN "Huh?"
-      1: RETURN "No way..."
+      OTHERWISE RETURN "Huh"
     ENDCASE
 ENDFUNCTION
 
--- a/main.pseudo
+++ b/main.pseudo
@@ -2,12 +2,14 @@ DECLARE Test : STRING
 
 Test <- "Hi"
 
-CASE OF Test
-  "Hello": OUTPUT "Oh, hello there!"
-  "Hi": OUTPUT "Hihi!"
-  "Hi": OUTPUT "Hi again!"
-  "Yo": OUTPUT "Wassup?"
-  0: OUTPUT "Huh?"
-  0: OUTPUT "Sense of deja vu ..."
-  1: OUTPUT "No way..."
-ENDCASE
+FUNCTION Respond(Prompt: STRING) RETURNS STRING
+    CASE OF Prompt
+      "Hello": RETURN "Oh, hello there!"
+      "Hi": RETURN "Hihi!"
+      "Yo": RETURN "Wassup?"
+      0: RETURN "Huh?"
+      1: RETURN "No way..."
+    ENDCASE
+ENDFUNCTION
+
+OUTPUT Respond(Test)
--- a/pseudocode/interpreter.py
+++ b/pseudocode/interpreter.py
@@ -10,14 +10,15 @@ from dataclasses import dataclass, field
 
 from . import (builtin, lang, system)
 
+
+
 # ----------------------------------------------------------------------
 
 # Helper functions
 
-
 def expectTypeElseError(exprmode: str,
                         *expected: str,
-                        errmsg: str = "Expected",
+                        errmsg: str="Expected",
                         token: lang.Token) -> None:
     """Takes in a type, followed by one or more expected types.
     Raises an error if the given type is not in the expected types.
@@ -25,27 +26,26 @@ def expectTypeElseError(exprmode: str,
     if not exprmode in expected:
         raise builtin.RuntimeError(f"{errmsg} {expected}", token=token)
 
-
 def declaredElseError(frame: lang.Frame,
                       name: lang.NameKey,
-                      errmsg: str = "Undeclared",
-                      token: Optional[lang.Token] = None) -> None:
+                      errmsg: str="Undeclared",
+                      token: Optional[lang.Token]=None) -> None:
     """Takes in a frame and a name.
     Raises an error if the name is not declared in the frame.
     """
     if not frame.has(name):
         raise builtin.RuntimeError(errmsg, token)
 
-
 def undeclaredElseError(frame: lang.Frame,
                         name: lang.NameKey,
                         errmsg="Already declared",
-                        token: Optional[lang.Token] = None) -> None:
+                        token: Optional[lang.Token]=None) -> None:
     """Takes in a frame and a name.
     Raises an error if the name is already declared in the frame.
     """
     if frame.has(name):
         raise builtin.RuntimeError(errmsg, token)
+
 
 
 @dataclass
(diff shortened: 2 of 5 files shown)


ngjunsiang:
Oops, I forgot I was handling leading line breaks earlier, and broke it with that commit. Putting it back in the scanner: [ef0e725272b3c4c7cbf63948c1de9ec4a5e88290]

ngjunsiang:
# Verifying RETURNs

Returns can be a bit tricky. Right now we are doing a naive check in the list of Procedure/Function statements:
https://github.com/nyjc-computing/pseudo-9608/blob/0d15d014fd2bfdb5ab83fe0fb0c01af5c0a7d68e/pseudocode/resolver.py#L393-L398
https://github.com/nyjc-computing/pseudo-9608/blob/0d15d014fd2bfdb5ab83fe0fb0c01af5c0a7d68e/pseudocode/resolver.py#L416-L419

ngjunsiang:
But suppose we have a function like this:
```
DECLARE Test : STRING

Test <- "Hi"

FUNCTION Respond(Prompt: STRING) RETURNS STRING
    CASE OF Prompt
      "Hello": RETURN "Oh, hello there!"
      "Hi": RETURN "Hihi!"
      "Yo": RETURN "Wassup?"
      0: RETURN "Huh?"
      1: RETURN "No way..."
    ENDCASE
ENDFUNCTION

OUTPUT Respond(Test)
```

This function contains only a CASE statement, no Returns; those are inside the case.Pseudo doesn't detect them:
```
$ pseudo main.pseudo 
[Line 5] FUNCTION Respond(Prompt: STRING) RETURNS STRING
                  ^
LogicError: [Line 5 column 10] 'Respond': No RETURN in function
```

ngjunsiang:
We are likely going to need a recursive solution. Currently, `verifyStmts()` takes an optional returnType argument. Let’s make it propagate this argument down to each `verify()` call, and have those propagate it to any `verifyStmts()` calls as well: [ecf0a67c2d1c8b293f51b5181556282f383c8788]

We'll have `verify()` handle any Return statements that it encounters, and handle Return statements through `verifyStmts()`, since that is what will be called through the handling of Function statements:
https://github.com/nyjc-computing/pseudo-9608/blob/ecf0a67c2d1c8b293f51b5181556282f383c8788/pseudocode/resolver.py#L352-L355

ngjunsiang:
We'll verify Return statements through `verifyStmts()`, since that is what will be called through the handling of Function statements: [5621dac34ebacb1859bea477b3c7143172009bb4]

ngjunsiang:
Time for another helper:

https://github.com/nyjc-computing/pseudo-9608/blob/ec8643b499dcc874c69f21c7f1558760f6421176/pseudocode/resolver.py#L314-L344

`willReturn()` takes a single Stmt and checks to see if it is guaranteed to Return. This is obvious True for a Return statement, and False for all other single-line statements.

For loops and conditionals we need to be a bit more careful.

In a loop, if any standalone statement (i.e. not nested in an IF, CASE, or other construct) is a Return, we take it that the loop will eventually return. We can't check if this will be an infinite loop or whether it will even run at all; that's a runtime check.

In a conditional, one of the cases might return, but the other cases might not. And if none of the cases match and there is no fallback, then there won't be a return at all. So we need to ensure that all branches have a Return before we return True from `willReturn()`.

ngjunsiang:
Okay, let's make it guarantee a return then:

[`main.pseudo`](https://github.com/nyjc-computing/pseudo-9608/blob/ec8643b499dcc874c69f21c7f1558760f6421176/main.pseudo)
```
DECLARE Test : STRING

Test <- "Hi"

FUNCTION Respond(Prompt: STRING) RETURNS STRING
    CASE OF Prompt
      "Hello": RETURN "Oh, hello there!"
      "Hi": RETURN "Hihi!"
      "Yo": RETURN "Wassup?"
      OTHERWISE RETURN "Huh"
    ENDCASE
ENDFUNCTION

OUTPUT Respond(Test)
```

```
$ pseudo main.pseudo 
[Line 10]       OTHERWISE RETURN "Huh"
                          ^
ParseError: [Line 10 column 17] 'RETURN': Unrecognised token
```

Oops .. something strange is going on. The problem is back in [`parser.py`](https://github.com/nyjc-computing/pseudo-9608/blob/ec8643b499dcc874c69f21c7f1558760f6421176/pseudocode/parser.py), on line 518:
https://github.com/nyjc-computing/pseudo-9608/blob/ec8643b499dcc874c69f21c7f1558760f6421176/pseudocode/parser.py#L502-L523

We are using `statement6()` to parse the OTHERWISE case, so it didn't pick up the RETURN. We should change it to `statement1()` to match the case parsing on line 511 ..

Actually, that doesn't sound right. Looking at the statement hierarchy, using `statement1()` will allow parsing of loops and conditionals within a case statement, which is explicitly against the 9608 pseudocode syntax. Anything more complicated than a single line should be handled with IF ... ELSE:

https://github.com/nyjc-computing/pseudo-9608/blob/ec8643b499dcc874c69f21c7f1558760f6421176/pseudocode/parser.py#L673-L687

ngjunsiang:
Hmm ... I can see a case (pun unintended) for putting RETURNs in CASE, which means ...another entry point for parsing statements:

https://github.com/nyjc-computing/pseudo-9608/blob/9bf6b75c8d133e816d7728572b74f902533446bf/pseudocode/parser.py#L686-L689
https://github.com/nyjc-computing/pseudo-9608/blob/9bf6b75c8d133e816d7728572b74f902533446bf/pseudocode/parser.py#L752-L755

And then we use it within the CASE statements: [9bf6b75c8d133e816d7728572b74f902533446bf]

```
  File "pseudocode/interpreter.py", line 197, in _
    return evalCallable(callable, expr.args, callable.frame)
  File "lib/python3.8/functools.py", line 875, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "pseudocode/interpreter.py", line 130, in _
    assert returnval, f"None returned from {callable}"
AssertionError: None returned from Function(frame={Prompt: STRING}, params=(<STRING: 'Hi'>,), stmts=[Case(cond=GetName(frame={Prompt: STRING}, name=Name(Prompt)), stmtMap={Literal(type='STRING', value='Hello', token=[Line 7 column 7] 'Hello'): [Return(expr=Literal(type='STRING', value='Oh, hello there!', token=[Line 7 column 23] 'Oh, hello there!'))], Literal(type='STRING', value='Hi', token=[Line 8 column 7] 'Hi'): [Return(expr=Literal(type='STRING', value='Hihi!', token=[Line 8 column 20] 'Hihi!'))], Literal(type='STRING', value='Yo', token=[Line 9 column 7] 'Yo'): [Return(expr=Literal(type='STRING', value='Wassup?', token=[Line 9 column 20] 'Wassup?'))]}, fallback=[Return(expr=Literal(type='STRING', value='Huh', token=[Line 10 column 24] 'Huh'))])])
```

Complicated message, but we know something: this happened in `interpreter.py`, which means it has successfully passed the parser and resolver stages.

And now, for some reason, the interpreter is failing to capture the return value from the CASE statements.

ngjunsiang:
Let's make it happen then:
https://github.com/nyjc-computing/pseudo-9608/blob/669fe827c4c0f8fa7f79c35e5e86db7f49af83ca/pseudocode/interpreter.py#L291-L329

Now our Loops and Conditionals check for a return value. If there's one, they assume it's from a RETURN statement, and halt execution there and then to pass the value along.

The rest of the commit is just formatting changes done by my linter: [669fe827c4c0f8fa7f79c35e5e86db7f49af83ca]

ngjunsiang:
The `execute()` dispatch is exactly the same for If and Case. We might as well merge them: [8f2dcf082960ac5cb72a5c17046ee822b5b6e4b5]

ngjunsiang:
One thing I don't like about the code right now:
https://github.com/nyjc-computing/pseudo-9608/blob/8f2dcf082960ac5cb72a5c17046ee822b5b6e4b5/pseudocode/interpreter.py#L131-L165

Those Unions tell me that my object type hierarchy is a mess. But that's more than I should be chewing off in this chapter, so let's close things here.

I forgot to mention that all unit tests passed, though `mypy` is almost certainly unhappy. Let's please it in another chapter.

PR #98, opened 2022-10-24
31c Defining behaviour: Object hierarchy

ngjunsiang:
This chapter, we begin clarifying what *things* we actually have in Pseudo, and how to classify them.

An unordered listing of things:
- Token
- Things we handle at evaluation: `bool`, `int`, `float`, `str`. Right now they are PyLiterals.
- Exprs and Stmts. Besides the subclasses of these, we also have some subcategories of these: SetExprs, ExprStmts, Conditionals, Loops, ProcFuncs, FileStmts
- Name, UnresolvedName (which is a subclass of Expr)
- Frame, TypeSystem
- TypedValue, TypeTemplate, ObjectTemplate
- Object, Array
- Builtin, Function, Procedure

I can't exactly see how to organise these just yet. Time to look through the parser, resolver, and interpreter, and see just hwo these are used.

Here's a first attempt:
Starting with the easy stuff: Let's make Builtins a Callable [cd0610baa3822654c416e2a05300dff6094d23af]

- [X] Token: These are pretty straightforward. They are used by the parser to determine the semantics of each token, and for error reporting.
- [X] **Exprs** and **Stmts**: We rely on these too much to change them now.
- [X] **Type**: strings representing valid Pseudo types, such as BOOLEAN, INTEGER, REAL, STRING, and any user-defined types.
- [X] **Name**: An identifier that points to a PseudoValue or Type
- [ ] **Indices**: Sequence of Exprs (evaluating to integers) that represent an array address
  - **Index**: A single Expr in **Indices**
- [ ] **PyLiterals**: This will remain. It's a convenient way to separate Python values from **PseudoValue**s.
  - `bool`
  - `int`
  - `float`
  - `str`
- [ ] **PseudoValue**: This is a convenient way to separate Pseudo objects from Python objects. **PseudoValue**s cover **Object**s, **Array**s, **Function**s/**Procedure**s/**Builtin**s, and **File**s. But we will need to subcategorise them.
  - [ ] **Callable**
    - Function
    - Procedure
    - Builtin
  - [ ] **Container**: This will encompass Objects and Arrays
  - [ ] **File**: This will stand alone by itself
- [X] **TypedValue**: This is a slot for storing PseudoValues in a TypedValueMap.
- [ ] **TypedValueMap**: Spaces that map Keys to **TypedValue**s
  - Frame: Maps **Name**s to **TypedValue**, can be chained to an outer **Frame**
  - Object: Maps **Name**s to **TypedValue**, cannot be chained
  - Array: Maps **Index** to **TypedValue**
- [ ] **Template**: These are not used directly to store values, but are used to clone TypedValueMaps
  - **TypeValueTemplate**: Used to clone **TypedValue**s. currently called TypeTemplate
  - **ObjectTemplate**: Used to clone **Object**s.


ngjunsiang:
Starting with the easy stuff: Let's make Builtins a Callable [cd0610baa3822654c416e2a05300dff6094d23af]

- [X] Token: These are pretty straightforward. They are used by the parser to determine the semantics of each token, and for error reporting.
- [X] **Exprs** and **Stmts**: We rely on these too much to change them now.
- [X] **Type**: strings representing valid Pseudo types, such as BOOLEAN, INTEGER, REAL, STRING, and any user-defined types.
- [X] **Name**: An identifier that points to a PseudoValue or Type
- [ ] **Index**: Sequence of Exprs (evaluating to integers) that represent an array address
- [ ] **PyLiterals**: This will remain. It's a convenient way to separate Python values from **PseudoValue**s.
  - `bool`
  - `int`
  - `float`
  - `str`
- [ ] **PseudoValue**: This is a convenient way to separate Pseudo objects from Python objects. **PseudoValue**s cover **Object**s, **Array**s, **Function**s/**Procedure**s/**Builtin**s, and **File**s. But we will need to subcategorise them.
  - [ ] **Callable**
    - Function
    - Procedure
    - Builtin
  - [ ] **Container**: This will encompass Objects and Arrays
  - [ ] **File**: This will stand alone by itself
- [X] **TypedValue**: This is a slot for storing PseudoValues in a TypedValueMap.
- [ ] **TypedValueMap**: Spaces that map Keys to **TypedValue**s
  - Frame: Maps **Name**s to **TypedValue**, can be chained to an outer **Frame**
  - Object: Maps **Name**s to **TypedValue**, cannot be chained
  - Array: Maps **Index** to **TypedValue**
- [ ] **Template**: These are not used directly to store values, but are used to clone TypedValueMaps
  - **TypeValueTemplate**: Used to clone **TypedValue**s. currently called TypeTemplate
  - **ObjectTemplate**: Used to clone **Object**s.

ngjunsiang:
And then let's rename Index and Indices: [c26aac6d52c4e66bf981f8fbf2ba4846f4a227f4]

Moving on to Containers: [d3a0502be1de1fcf1595acc76304d09a3a492261]

Annoyingly, `mypy` doesn't seem to recognise the forward definition for Container, so we have to move it to after the class definitions for Container, Array, and Object. We'll figure it out later; seems to be a cyclic import issue.

### Progress
- [X] Token: These are pretty straightforward. They are used by the parser to determine the semantics of each token, and for error reporting.
- [X] **Exprs** and **Stmts**: We rely on these too much to change them now.
- [X] **Type**: strings representing valid Pseudo types, such as BOOLEAN, INTEGER, REAL, STRING, and any user-defined types.
- [X] **Name**: An identifier that points to a PseudoValue or Type
- [X] **Indices**: Sequence of Exprs (evaluating to integers) that represent an array address
  - **Index**: A single Expr in **Indices**
- [ ] **PyLiterals**: This will remain. It's a convenient way to separate Python values from **PseudoValue**s.
  - `bool`
  - `int`
  - `float`
  - `str`
- [ ] **PseudoValue**: This is a convenient way to separate Pseudo objects from Python objects. **PseudoValue**s cover **Object**s, **Array**s, **Function**s/**Procedure**s/**Builtin**s, and **File**s. But we will need to subcategorise them.
  - [X] **Callable**
    - Function
    - Procedure
    - Builtin
  - [X] **Container**: This will encompass Objects and Arrays
    - Object
    - Array
  - [ ] **File**: This will stand alone by itself
- [X] **TypedValue**: This is a slot for storing PseudoValues in a TypedValueMap.
- [ ] **TypedValueMap**: Spaces that map Keys to **TypedValue**s. Since **Object**s and **Array**s are already subclasses of **Container**s, this will be a TypeAlias if necessary.
  - Frame: Maps **Name**s to **TypedValue**, can be chained to an outer **Frame**
  - Object: Maps **Name**s to **TypedValue**, cannot be chained
  - Array: Maps **Index** to **TypedValue**
- [ ] **Template**: These are not used directly to store values, but are used to clone TypedValueMaps
  - **TypeValueTemplate**: Used to clone **TypedValue**s. currently called TypeTemplate
  - **ObjectTemplate**: Used to clone **Object**s.

ngjunsiang:
Now we try to tame the bucking horse that is Objects and Arrays, resolving mypy's complaints one by one.

Central to this is the Assignable type, which we use to represent things that can be assigned in an assignment expression: PyLiterals, Array Indexes, and Object Attributes. This is typed as Union[PyLiteral, Container]. And we use this to type Objects, as well as return values: [24df2b7643ac1174309f70f20eeb5890dd007bcb]

It’s ugly, but it works, and we'll make it neater when we rethink Frames later.

ngjunsiang:
Finally we implement the Template parent class. Here we throw in an abstract base class (ABC), to indicate that Template should never be directly instantiated. And we also define a Template as a class having a `clone()` method, by using the `@abstractmethod` decorator: [e5411a15953deee3eb4a9f38fe5bfab4b7f12ab9]

https://github.com/nyjc-computing/pseudo-9608/blob/e5411a15953deee3eb4a9f38fe5bfab4b7f12ab9/pseudocode/lang.py#L132-L144

### Progress
- [X] Token: These are pretty straightforward. They are used by the parser to determine the semantics of each token, and for error reporting.
- [X] **Exprs** and **Stmts**: We rely on these too much to change them now.
- [X] **Type**: strings representing valid Pseudo types, such as BOOLEAN, INTEGER, REAL, STRING, and any user-defined types.
- [X] **Name**: An identifier that points to a PseudoValue or Type
- [X] **Indices**: Sequence of Exprs (evaluating to integers) that represent an array address
  - **Index**: A single Expr in **Indices**
- [X] **PyLiterals**: This will remain. It's a convenient way to separate Python values from **PseudoValue**s.
  - `bool`
  - `int`
  - `float`
  - `str`
- [X] **PseudoValue**: This is a convenient way to separate Pseudo objects from Python objects. **PseudoValue**s cover **Object**s, **Array**s, **Function**s/**Procedure**s/**Builtin**s, and **File**s. But we will need to subcategorise them.
  - [X] **Callable**
    - Function
    - Procedure
    - Builtin
  - [X] **Container**: This will encompass Objects and Arrays
    - Object
    - Array
  - [X] **File**: This will stand alone by itself
- [X] **TypedValue**: This is a slot for storing PseudoValues in a TypedValueMap.
- [ ] ~~**TypedValueMap**: Spaces that map Keys to **TypedValue**s. Since **Object**s and **Array**s are already subclasses of **Container**s, this will be a TypeAlias if necessary.~~
- [X] Frame: Maps **Name**s to **TypedValue**s, can be chained to an outer **Frame**
- [X] **Template**: These are not used directly to store values, but are used to clone TypedValueMaps
  - **TypeValueTemplate**: Used to clone **TypedValue**s. currently called TypeTemplate
  - **ObjectTemplate**: Used to clone **Object**s.

We are almost done! Looking at the todo list, and at the types that are being used, I'm rethinking the utility of TypedValueMap; nowhere in the code do we need to differentiate Frames from Objects and Arrays. LEt's take it out for now.

ngjunsiang:
Part of the problem is that `lang.py` is getting somewhat unwieldy: we have PseudoValues, Exprs, and Stmts all lumped into one file, and all depending on each other. This was necessary at the start because these types can be somewhat recursive, but if we want to have hierarchical imports we have to have a hierarchy!

Let's start by cleaning up: [f3ed5cb2a37a9b82bd8105a1141143df449cde84]

ngjunsiang:
I start by splitting off TypeSystem and Frame, pulling with it all its dependencies: [56c036435697d282441e2581675f6cd5abdb34ca]

This gives us a kind of nice split between PseudoValues in `object.py`, and Exprs/Stmts + their dependencies.

ngjunsiang:
We might as well pull Frame over: [a94d43e5edb3e3df7fafe061be641517de198060], [5bd0c29379a170352d6c088d9f45cd541aa66146]

ngjunsiang:
I'm seeing some questionable dependencies in our Containers ... can someone explain why Arrays and Objects need a TypeSystem?

The thing about having a commit history (and detailed pull requests like in this book) is I can find the exact commit when this happened: https://github.com/nyjc-computing/pseudo-9608/pull/58#issuecomment-1128590485

At one point I tried to make Frame and Object polymorphic. That turned out to not fit the implementation of Pseudo very well, with Frame no longer inheriting from PseudoValue. So we can undo that now ... but wait! How will we initialise the TypedValues inside the Array?

https://github.com/nyjc-computing/pseudo-9608/blob/5bd0c29379a170352d6c088d9f45cd541aa66146/pseudocode/lang/object.py#L306-L313

The critical codepath is here:

https://github.com/nyjc-computing/pseudo-9608/blob/5bd0c29379a170352d6c088d9f45cd541aa66146/pseudocode/resolver.py#L118-L130

In the resolver, we use the Array's typesystem to clone any declared objects. The Array itself doesn't need a type system otherwise; note that the typesystem comes from the frame anyway, which should have access to the typesystem (through its `types` attribute).

ngjunsiang:
The code is still working at this point; I got psyched out and didn't start using the TypeSystem instantiated in Pseudo. It's just sitting there unused!

https://github.com/nyjc-computing/pseudo-9608/blob/7d5b5cd139f1884504b8ce7be20e406e08ce301a/pseudocode/__init__.py#L94-L98

There it is, the `typesys` attribute. It is passed to Resolver and Interpreter:

https://github.com/nyjc-computing/pseudo-9608/blob/7d5b5cd139f1884504b8ce7be20e406e08ce301a/pseudocode/__init__.py#L143-L164

And Resolver is ... doing nothing with it:

https://github.com/nyjc-computing/pseudo-9608/blob/7d5b5cd139f1884504b8ce7be20e406e08ce301a/pseudocode/resolver.py#L95-L103

Let's think about what is going to change: all the `verify*()` helpers will have to take an extra argument in addition to `frame`. And when we do more decoupling (I just know it's coming), it's more arguments ... what a mess!

For a start, we have `frame`, and about to add `typesys`. I have a good mind to separate `function`s into a separate frame too, since functions are not first-class in 9608 pseudocode and should not go in the same place as variables. And later, if we incorporate constants into `lang.py`, we might need to add a Frame for constants too. That's at least 4 arguments to pass around the resolver and verifier helpers.

Let's bundle them into something ... we'll call it the Context? No, it's more like ... an Environment!

ngjunsiang:
Just small changes first. Meet Environment:

https://github.com/nyjc-computing/pseudo-9608/blob/4e8c444275a31fc56da8ce7217e9d2bb7df1d2a5/pseudocode/__init__.py#L38-L48

I'm commenting out `typesys` first because I'm going to break the existing code slightly in order to introduce Environment to the rest of the interpreting pipeline. I'll be trying to get it back in working order, and the fewer complications involved, the easier.

ngjunsiang:
We're keeping some redundancy here; `env` has a reference to frame, and we're not doing away with `self.frame` just yet.

https://github.com/nyjc-computing/pseudo-9608/blob/be668d0de350fcd0b6cb6bed1b846f300c970e6f/pseudocode/__init__.py#L104-L109

It gets passed to Resolver and Interpreter here:
https://github.com/nyjc-computing/pseudo-9608/blob/be668d0de350fcd0b6cb6bed1b846f300c970e6f/pseudocode/__init__.py#L154-L155
https://github.com/nyjc-computing/pseudo-9608/blob/be668d0de350fcd0b6cb6bed1b846f300c970e6f/pseudocode/__init__.py#L165-L166

Which receive them in `resolver.py` and `interpreter.py`:
https://github.com/nyjc-computing/pseudo-9608/blob/be668d0de350fcd0b6cb6bed1b846f300c970e6f/pseudocode/resolver.py#L100-L105
https://github.com/nyjc-computing/pseudo-9608/blob/be668d0de350fcd0b6cb6bed1b846f300c970e6f/pseudocode/interpreter.py#L56-L62

Okay, bad idea to put Environment in `__init__.py`; I initially thought it should not be considered part of the language. But clearly the resolver and interpreter will have to import it to know about its type behaviour, and they can't import it from `__init__.py` since *they are imported by init*!

ngjunsiang:
Okay, huge changes incoming: [2a6e6f6e4d0e6659d9007f81b305c744ce95df12]

This is where we carefully trawl through `resolver.py` and `interpreter.py`, changing all the resolving and verifying helpers to take an Environment instead of a frame, and making sure that every call is passed an `env` instead of a `frame`.

We're not 100% done though. There are a couple of tricky places:
https://github.com/nyjc-computing/pseudo-9608/blob/2a6e6f6e4d0e6659d9007f81b305c744ce95df12/pseudocode/resolver.py#L429-L470

https://github.com/nyjc-computing/pseudo-9608/blob/2a6e6f6e4d0e6659d9007f81b305c744ce95df12/pseudocode/interpreter.py#L124-L142
https://github.com/nyjc-computing/pseudo-9608/blob/2a6e6f6e4d0e6659d9007f81b305c744ce95df12/pseudocode/interpreter.py#L225-L230
https://github.com/nyjc-computing/pseudo-9608/blob/2a6e6f6e4d0e6659d9007f81b305c744ce95df12/pseudocode/interpreter.py#L400-L403

Basically these are places where Procedures and Functions are declared, have a local frame instantiated, and then these frames are stored in Callables, to be used by the interpreter.

Unfortunately for all involved, the resolvers and verifiers no longer take frames, but environments. That means ... we'll need a new environment for each callable?!

An environment is just a shell, holding on to references to the global frame, and soon the typesystem. Creating a new environment doesn't mean also duplicating the frame and typesystem. So it's no issue really, but we'll have to swap in the local frame in place of the global frame in the new environment.

We could do something like:

```
localenv = lang.Environment(frame=env.frame, typesys=env.types, ...)
```

which looks pretty ugly. Let's use an instance method to do something like the following:

```
localenv = env.with_frame(local)
```

Which should return a new environment with the same typesystem and everything else, and only the frame swapped out for the local frame. (Remember that the local frame still holds its own reference to the global frame through its `outer` attribute, so it is stil able to resolve global variables.)

ngjunsiang:
We had to do some major surgery, starting with the `with_frame()` method:
https://github.com/nyjc-computing/pseudo-9608/blob/57ec25233a9d5ef3889d4d588ecf40b14467e05c/pseudocode/lang/__init__.py#L114-L116

We wrap the local frame in an environment before using it:
https://github.com/nyjc-computing/pseudo-9608/blob/57ec25233a9d5ef3889d4d588ecf40b14467e05c/pseudocode/resolver.py#L437-L446
https://github.com/nyjc-computing/pseudo-9608/blob/57ec25233a9d5ef3889d4d588ecf40b14467e05c/pseudocode/resolver.py#L457-L472

Why do we need to change the Callable frames to envs? Because of these lines in `interpreter.py`:

https://github.com/nyjc-computing/pseudo-9608/blob/57ec25233a9d5ef3889d4d588ecf40b14467e05c/pseudocode/interpreter.py#L130
https://github.com/nyjc-computing/pseudo-9608/blob/57ec25233a9d5ef3889d4d588ecf40b14467e05c/pseudocode/interpreter.py#L140
https://github.com/nyjc-computing/pseudo-9608/blob/57ec25233a9d5ef3889d4d588ecf40b14467e05c/pseudocode/interpreter.py#L229

Here we used to pass the frames for execution, but the executing helpers now take envs instead, so everything has to follow suit.

That means Builtin callables now have to be instantiated with Environments too, necessitating changes in `system.py`:

https://github.com/nyjc-computing/pseudo-9608/blob/57ec25233a9d5ef3889d4d588ecf40b14467e05c/pseudocode/system.py#L154-L163

You can see the carnage in the full commit: [57ec25233a9d5ef3889d4d588ecf40b14467e05c]

It would have been difficult to make any of these changes in isolation and still have working code. But at least that's over and we're back to all tests passing! (We did have to fix one test that assumed procedures still used frames).

ngjunsiang:
Are we ready to stop passing frames in Pseudo? I commented out those lines and ... broke a whole bunch of tests that relied on the Result object having a frame. Time to convert those tests, since Result now has the full environment: [530302f8afc18cceb7183e85fed332a304ad8ffd]

ngjunsiang:
The only place that uses the typesystem is the resolver. Easy peasy, find-and-replacey: [2015f0b127810e94ca3078f48b13ef9f176ba9b7]

The typesystem we are using is still being initialised in `system.py` though. I want it initialised in Pseudo. Let's do that, and pass the typesystem to `system.initFrame()` to use first: [5ddce22b357745a03bbbd5ed7bffa7375d2a418a]

ngjunsiang:
We are still reliant on the typesystem in Frame and Object, and can't remove them yet, because they are being used in the `declare()` method:
https://github.com/nyjc-computing/pseudo-9608/blob/5ddce22b357745a03bbbd5ed7bffa7375d2a418a/pseudocode/lang/object.py#L225-L226

We are going to take that away from Frame; it will have no notion of typesystems. When it declares a type, it will have to take the cloned TypeValue through the `declare()` method, since when we are done it will no longer have a reference to the typesystem.

[f07a78eb7302a889309326185f0d68a20c7cbbf3]

This is a breaking change: Frame/Object declarations won't work until we fix `resolver.py`.

ngjunsiang:
That was easier than expected: replace all `declare()` calls and make them take a cloned TypedValue from the typesystem instead of trying to do the cloning themselves.

[f045d7549dde144d32d07c35de62f47c31a418e5]

We are back to working code!

ngjunsiang:
Time to take typesystem away from Containers ... how to do this? I started by commenting out the `types` initialisation in the constructor, and quickly found out:
https://github.com/nyjc-computing/pseudo-9608/blob/f045d7549dde144d32d07c35de62f47c31a418e5/pseudocode/lang/object.py#L319-L326

Arrays are still using their typesystem in their constructor! Time to move that out to the resolver: [58c5dea0ea696636c61945160dc03f6aa56e9291]

ngjunsiang:
Now we sneakily remove typesystem from Frame and Containers, and ... also make minor edits to Pseudo's `__init__()` plus fix a couple of old typos in the resolver: [9154ecd194c09f4114125c153cdeaa6a626604ba]

Code's still working ...

ngjunsiang:
Now we make breaking changes to the constructor of Frame and Containers, not even passing in any typesystem. And just as quickly we fix them, changing each call to Frame, Object, and Array: [1495570ef142b94301b0d6a75fe0ba0dd239137a]

ngjunsiang:
We did it!: [https://github.com/nyjc-computing/pseudo-9608/pull/98/commits/cc24e00163cfe04f38d2636f01f51e8a1aa35c2d]

Now we have a clear import hierarchy:
`types` --> `object`
`types` --> `typesystem`
`object` --> `typesystem`

And all three export their namespace into `lang.py`, so that we don't need to scratch our heads whether to import a type from `lang.types`, `lang.object`, or `lang.typesystem`. This decoupling is to help us keep things straight!

ngjunsiang:
I've achieved what I set out to do; while there's still more to be done, that's enough for this chapter.

